Issue #3, opened 2016-07-20
An idea about 'nc_rpc_clb'


fanchanghu:
Hi,
I have met a question about handling RPC request. I want to use one common callback for all my RPC requests (in my case, just dispatch all RPC request to sysrepo with `sr_rpc_send`). But now I must set the callback function pointer to lys_node for all RPC, so I think we should allow users to give a global callback for handling RPC request, and if this global callback is not set, we can get callback function from lys_node to make sure current API still works.
If you agree with this idea, I can work on it and push the code to you.

PR #4, opened 2016-07-21
add global RPC callback feature

fanchanghu:
Hi,
I discussed this feature in Issue #3, and I have made a little change about it: first try the private field in the corresponding RPC schema node, if it is null, use the global RPC callback, and if it is also null, return an error.


rkrejci:
That "little change" is exactly what I wanted to propose to you ;)

PR #6, opened 2016-07-26
Improvements to CMake when LibSSH is not found

jktjkt:
...by switching from a handcrafted module to pkg-config.
    
With the original cmake 2.8.11 on CentOS 7.2.1511, a missing LibSSH was
not reported through a usual "required package XYZ not found" message,
but it generated a bit unintuitive error:
    
      CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
      Please set them or make sure they are set and tested correctly in the CMake files:
      LIBSSH_INCLUDE_DIR
         used as include directory in directory /home/centos/libnetconf2
     (...)
         used as include directory in directory /home/centos/libnetconf2/tests
    
I suspect that this happens because the custom FindLibSSH.cmake doesn't
set LIBSSH_FOUND to TRUE on success (on my system).
    
According to rpmfind.net [1], the required pkg-config files have been
shipping since at least the release which is available in EL6's EPEL.
    
[1] https://www.rpmfind.net/linux/rpm2html/search.php?query=pkgconfig(libssh)


michalvasko:
Hi,
if you have the correct libssh package installed then the *FindLibSSH.cmake* module should definitely find it. The error may not be perfect, but it should be obvious what the problem is. The CMake module is written by the libssh team and I don't really see a reason why not to use it.

Regarding skipping -l in the additional linked libraries, are you saying it did not work for you the way it is?

Regards,
Michal

jktjkt:
> if you have the correct libssh package installed then the FindLibSSH.cmake module should definitely find it

Correct; there's no problem if that library is already installed.

> The CMake module is written by the libssh team and I don't really see a reason why not to use it.

It doesn't conform to CMake's conventions. A ```FindFoo.cmake``` should always set ```Foo_FOUND``` to ```TRUE``` if that library is available. This bundled file does not set that variable on success -- it only sets it upon failure, which is why the ```find_package(... REQUIRED)``` doesn't error out early, and why the ```cmake``` process only gets interrupted by trying to use an undefined variable. As you said, a file coming form libssh means that this is a libssh bug. As this case demonstrates, a ```pkg-config``` approach would eliminate this error. The decision is yours for sure.

> Regarding skipping -l in the additional linked libraries, are you saying it did not work for you the way it is?

Nope, it works already. I changed it because I was investigating another issue, it's just a followup cleanup that I split from my original commit (unpushed) for clarity. Feel free to ignore that one if you don't like it -- I was removing that hardcoded and redundant ```-lssh_threads``` so I cleaned up the rest as well.

michalvasko:
Hi,
it seems to me all this just polishes the build script, but does not fix any actual errors. So thanks for the effort, I will consider all this in the future when I have time for such minor fixes of the whole project, but I will not merge it now.

Regards,
Michal

PR #13, opened 2016-11-03
Build with LibSSH 0.6.3, too

jktjkt:
...because that's what is available in Debian Jessie.

### -> Is this OK from your point of view?

Maybe there are some important bugfixes/features in that release which I'm not aware of. Changelog suggests new support for ECs, etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,7 +112,7 @@ check_function_exists(pthread_mutex_timedlock HAVE_PTHREAD_MUTEX_TIMEDLOCK)
 
 # dependencies - libssh
 if(ENABLE_SSH)
-    find_package(LibSSH 0.6.4 REQUIRED)
+    find_package(LibSSH 0.6.3 REQUIRED)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DNC_ENABLED_SSH ${LIBSSH_DEFINITIONS}")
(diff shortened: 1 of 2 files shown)


rkrejci:
I'm afraid that there really was some bug that appeared already in libnetconf. However, this is question for @michalvasko (he's currently not avavilable, should be back online in 10 days), so, please, let's wait for his answer.

jktjkt:
Thanks, I've already noticed a difference about handling of ``SSH_BIND_OPTIONS_HOSTKEY`` (see the updated commit).

Anyway, this can wait, sure.

michalvasko:
Hi,
libssh 0.6.3 does not support ECDSA keys, for one, if I remember correctly. However, the reason we require 0.6.4 or higher is that in 0.6.3 was a communication-breaking bug when using own sockets, no message could be received I think and the reading hanged forever.

Regards,
Michal

abhikeshav:
Hi @michalvasko , I'm running across some issue when using libssh 0.6.3 with libnetconf2/libnetconf where the connection is hanging. Just saw your comment. Is this only issue only when using own sockets or are there problems using libssh 0.6.3?

Thanks a lot in advance for your time!

michalvasko:
Hi,

in *libnetconf2* we always use own sockets, so that is why I encountered it. Whether there are any other problems I don't know.

Regards,
Michal

PR #22, opened 2017-03-13
session BUGFIX capabilitys do not include ns if it is prefix of another

fanchanghu:
Hi,
I found the namespace  is not sent to client in hello message, if it is prefix of another one. Such as "uri:ab" will not be added if "uri:abc" has already be added.


michalvasko:
Hi,
thanks, seems fine.

Regards,
Michal

PR #36, opened 2017-08-01
session BUGFIX block forever

fanchanghu:
Hi,
When a client quits before sending hello message, the server will block forever if its hello timeout is not set. This request try to fix it by setting the default hello timeout to 60s in that situation.


michalvasko:
Hi,
seems reasonable and you nicely kept style with the rest of the code, thanks.

Regards,
Michal

PR #37, opened 2017-08-03
server session BUGFIX where crypt_r is not available

mislavn:
Hi,

I encountered a portability issue with uclibc in OpenWrt. This patched fixed my problem and it has been tested on an older OpenWrt build (uclibc), on the latest Lede build (musl) and in docker (ubuntu 16.04).

### DIFF
--- a/src/session_server_ssh.c
+++ b/src/session_server_ssh.c
@@ -28,6 +28,10 @@
 #include "session_server_ch.h"
 #include "libnetconf.h"
 
+#if !defined(HAVE_CRYPT_R)
+pthread_mutex_t crypt_lock = PTHREAD_MUTEX_INITIALIZER;
+#endif
+
 extern struct nc_server_opts server_opts;
 
 static char *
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
that seems fine except I would have initially used `crypt` if it was correct. With your patch if 2 user get to authenticate at the exact same time and `crypt` gets called simultaneously, it will likely crash. It would be better to think about some other solution that will not break thread-safety.

Regards,
Michal

mislavn:
Hi,

@michalvasko is it enough if I add locks to the function auth_password_compare_pwd() just for the case when crypt_r is not defined?

Regards,
Mislav

michalvasko:
Hi Mislav,
yeah, that should work fine for cases where `crypt_r` cannot be used.

Regards,
Michal

michalvasko:
Hi,
seems ok except I don't see any initialization of the lock, so just initialize it statically and it should be ready to be merged.

Regards,
Michal

mislavn:
Done.

PR #38, opened 2017-08-09
server BUGFIX nc_ps_poll deadlock due to multi-thread conflict

fanchanghu:
Hi,
`nc_ps_poll` will enter a deadlock branch in the line 1412, this issue can be reproduced in following steps:
1, current session is in the last position of `ps->sessions`;
2, state of current session is been set to `NC_PS_STATE_BUSY`; 
3, current thread unlocks the ps lock in the line 1469;
4, another session is removed from  `ps->sessions` by another thread, and current session is copied to the removed position;
5, current thread set state of current session (in the old position, which is invalid now) to `NC_PS_STATE_NONE`;
6, the copied session state is always `NC_PS_STATE_BUSY`;
7, threads trap into deadlock when they handle this copied session.

By changing the `sessions` in the `struct nc_pollsession` to a pointer array, we can just move a pointer but not a struct when removing a session, then this issue is fixed.

### DIFF
--- a/src/session_p.h
+++ b/src/session_p.h
@@ -424,12 +424,14 @@ enum nc_ps_session_state {
     NC_PS_STATE_INVALID        /**< session is invalid and was already returned by another poll */
 };
 
+struct nc_ps_session {
+    struct nc_session *session;
+    enum nc_ps_session_state state;
+};
+
 /* ACCESS locked */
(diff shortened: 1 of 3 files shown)


michalvasko:
Hi,
nice catch and well-formatted patch, thanks!

Regards,
Michal

PR #41, opened 2017-09-15
BUGFIX - fix a potential crash in ERR prints.

darshana-jk:


michalvasko:
Thanks, merged.

Regards,
Michal

PR #45, opened 2017-10-16
session BUGFIX segfault and corrupt server capabilities string with deviations

hanneskuettner:
Hi,
I encountered a corrupt server capabilities string when using deviated modules. 
This has to do with an incorrect parse of the deviation node of the [RFC 7895](https://tools.ietf.org/html/rfc7895) `ietf-yang-library` data set. 
The segfault was due to allocating space for the node but never actually assigning it.

And instead of the correct usage of deviations as `common-leafs` it was assumed that it is a list of `yang:yang-identifier` as the feature set. This lead to garbage output when the server has module deviations loaded. 

Fixed by correctly parsing the deviations data node.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -881,6 +881,7 @@ nc_server_get_cpblts(struct ly_ctx *ctx)
                         ERRMEM;
                         goto error;
                     }
+                    deviations[dev_count - 1] = (struct lyd_node_leaf_list *)child2;
                 }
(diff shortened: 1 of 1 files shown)


hanneskuettner:
Any comments?

michalvasko:
Right, thank you.

Regards,
Michal

PR #49, opened 2018-02-14
Use proper size for inet_ntop's buffer

jktjkt:
The manpage suggests to use this macro. The old code was failing for IPv6-mapped IPv4 addresses.


michalvasko:
Hi,
you are right, thanks.

Michal

PR #52, opened 2018-03-21
Prevent the socket fd leak:

darshana-jk:
Close the socket when SSL session cleanup.
Fix for: https://github.com/CESNET/libnetconf2/issues/53


michalvasko:
Hi Darshana,
you are getting the socket from a freed structure, I will fix it.

Regards,
Michal

PR #55, opened 2018-03-28
client session BUGFIX SSH passwd auth segfault

leonshaw:
Hi,
After su from root to a normal user, ttyname of stdin is usually /dev/ttyx or /dev/pts/x, which is owned by root with permission 0620. The user has insufficient permission to the TTY resulting null password. It should use /dev/tty rather than getting ttyname from stdin.

### DIFF
--- a/src/session_client_ssh.c
+++ b/src/session_client_ssh.c
@@ -57,7 +57,7 @@ open_tty_noecho(const char *path, struct termios *oldterm)
     struct termios newterm;
     FILE *ret;
 
-    if (!(ret = fopen(path, "r"))) {
+    if (!(ret = fopen(path, "a+"))) {
         ERR("Unable to open the current terminal (%s).", strerror(errno));
         return NULL;
     }
@@ -332,7 +332,7 @@ char *
 sshauth_password(const char *username, const char *hostname, void *UNUSED(priv))
 {
     char *buf;
-    int buflen = 1024, len, ret;
+    int buflen = 1024, len;
     char c = 0;
     struct termios oldterm;
     FILE *tty;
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
your pull request is generally fine but I will not merge it as it is. Instead, first a terminal will be attempted to be opened the way it is now and if it fails then `/dev/tty` will be used as a backup. We initially had it implemented exactly the way you propose but it did not work in cases where `stdout` and `stdin` were not the same terminal/devices.

Before I make the change I want to ask why have you changed the mode for `fopen` from `r` to `a+`?

Regards,
Michal

michalvasko:
Hi,
what do you think about it now? It worked for me having previously reproduced your problem.

Regards,
Michal

leonshaw:
@michalvasko 

Hi,

Functionally it's fine, but I think this is redundant. If stdin is a terminal, the result is the same, but if stdin is not a terminal, ttyname_r may still fail. Using /dev/tty for both input and output also allows redirecting while being able to ask password interactively. For example with netopeer2:
`netopeer2-cli < cmds > results`

Thanks,
Xiao

michalvasko:
Hi Xiao,
as long as it works for you, we will keep it this way because like I said, it is not redundant. `netopeer2-cli` scripting is already possible.

Regards,
Michal

leonshaw:
@michalvasko 

But the example above will give an error:
```
# netopeer2-cli < cmds > results
nc ERROR: ttyname_r failed (Inappropriate ioctl for device).
Segmentation fault (core dumped)
```
cmds:
```
connect --ssh --host 127.0.0.1 --login root
status
```

BTW, I'm actually referencing the implementation of openssh.

michalvasko:
Hi,
it should no longer crash with the referenced fix in `devel` and I have briefly described how to use scripts [here](http://lists.sysrepo.org/archives/sysrepo-devel/2018-January/000308.html). I agree that is not as straightforward as your example but can be successfully used nonetheless.

We could switch it and try `/dev/tty` first, that is true but I do not think it matters. It is attempted to use `stdin` for situations when `/dev/tty` is not available or it is not the same as `stdin` (rare situations, but we have encountered them).

Regards,
Michal

leonshaw:
@michalvasko 
Hi,

I don't see code change handling failure of ttyname_r.

It's easy to create such situation. Suppose we are on /dev/tty1:
`netopeer2-cli < /dev/tty2 > /dev/tty3`

1. Using /dev/tty, the prompt and password is printed/read on /dev/tty1.
2. Using stdin and stdout, the prompt is printed on /dev/tty3, and password is read from /dev/tty2.

I prefer the behavior of 1, which is the same as openssh client. Generally people would expect inputing password on the TTY where they run the program, which is the controlling terminal. It's more probably that stdin is not a terminal than that /dev/tty is not available.
If you still insist on using stdin, at least the prompt and password input should be on the same TTY.

Thanks,
Xiao

michalvasko:
HI Xiao,
I hope you are satisfied now, everything should work like you wanted.

Regards,
Michal

leonshaw:
That's good. Thanks!

PR #58, opened 2018-04-16
server BUGFIX.

darshana-jk:
Properly handle call-home endpoint deletions & memory reallocation.
Prevent server from accessing a dangling pointer when call-home endpoints deleted/replaced.

We experienced a segfault in server when the full list of call home entries are replaced with new ones. gdb showed that the CH thread access memory which is not accessible (cur_endpt assumes that the memory is there, but its being moved). The changes handle call-home endpoint deletion properly. Also added proper accessing to the 'cur_endpt' pointer. Pls review.

### DIFF
--- a/src/session_server.c
+++ b/src/session_server.c
@@ -2283,15 +2283,33 @@ nc_server_ch_client_del_endpt(const char *client_name, const char *endpt_name)
                 lydict_remove(server_opts.ctx, client->ch_endpts[i].name);
                 lydict_remove(server_opts.ctx, client->ch_endpts[i].address);
 
-                /* move last endpoint to the empty space */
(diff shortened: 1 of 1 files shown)


rkrejci:
Hi, can you be more specific about that segfault. I don't see what your patch is actually solving. The access to the removed items in the list should be limited via `ch_endpt_count` and it seems correct to me. What is the problem you are solving by the patch? If I understand it correctly, you just shrink the memory allocated for the array, right? But why? When `ch_endpt_count` is correctly used, the program should not access the items that are no more active, should it?

michalvasko:
Hi Darshana,
it works now, I hope.

Regards,
Michal

PR #61, opened 2018-06-05
Add optional authentication methods for ssh pubkey and interactive.

bhart3:
Hey everyone,

We wanted to add optional callback functions for pubkey and interactive authentication for ssh. We noticed you had implemented it for password but we added functionality that will also authenticate users differently for pubkey and interactive if you register callbacks. We modeled them after the existing changes to password authentication. 

Please let us know your thoughts.

Thanks,
Brandon Hart

### DIFF
--- a/src/session_server.h
+++ b/src/session_server.h
@@ -22,6 +22,12 @@
 #   include <openssl/x509.h>
 #endif
 
+#ifdef NC_ENABLED_SSH
+#   include <libssh/libssh.h>
+#   include <libssh/callbacks.h>
+#   include <libssh/server.h>
+#endif
+
 #include "session.h"
 #include "netconf.h"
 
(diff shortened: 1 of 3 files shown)


bhart3:
Sorry about the formatting errors. We attempted to keep the formatting the same but we definitely missed some of them. 

We will make the changes that you suggested and re-upload a new version soon. Thank you very much for the feedback!

Regards,
Brandon

bhart3:
One question about the comment on passing the user that is trying to authenticate into the callback functions. Can we not just use the session struct to grab the username? Which I think is what password authentication is doing as well.

src/session_server.h
```diff
@@ -521,7 +521,7 @@ void nc_server_ssh_set_passwd_auth_clb(int (*passwd_auth_clb)(const struct nc_se
  * @param[in] user_data Optional arbitrary user data that will be passed to \p passwd_auth_clb.
  * @param[in] free_user_data Optional callback that will be called during cleanup to free any \p user_data.
  */
-void ncserver_ssh_set_interactive_auth_clb(int (*interactive_auth_clb)(const struct ssh_message msg, const char *password,
+void ncserver_ssh_set_interactive_auth_clb(int (*interactive_auth_clb)(const struct nc_session *session, const struct ssh_message msg,
```
bhart3: I have changed the callback function to accept the message that was sent, so that the callback function can parse the interactive responses and respond appropriately.

bhart3:
Let me know if this seems more inline with your suggestions so that the questions and responses for the interactive authentication can be handled by the callback function that registers.

michalvasko:
Hi,
should be fine now, I have just corrected some formatting errors. Thanks.

Regards,
Michal

PR #62, opened 2018-06-12
Fix segfault when crypt() returns NULL

alangefe:
`crypt()` and `crypt_r()` can return `NULL` when an error is encountered. For example, if a user has been locked using `usermod -L`, queries to the password database can return an encrypted password prefixed by `!`, and `crypt()` will return `NULL` because the password hash is malformed.

This change prevents `auth_password_compare_pwd()` from dereferencing a NULL pointer if `crypt()` or `crypt_r()` returns NULL due to an error. Instead, we now return nonzero in this case (i.e., authentication failed).


michalvasko:
Okay, no problem.

Regards,
Michal

PR #66, opened 2018-07-09
session: do not free rpc request external data

rjarry:
When making a generic rpc request:

```c
struct lyd_node *input;
struct nc_reply *reply
struct nc_rpc *rpc;
uint64_t msgid;

input= lyd_new_path(NULL, ctx, "/example:my-rpc/parameter", "value",
                    LYD_ANYDATA_CONSTSTRING, LYD_PATH_OPT_UPDATE);

rpc = nc_rpc_act_generic(input, NC_PARAMTYPE_CONST);
nc_send_rpc(session, rpc, 1000, &msgid);
nc_recv_reply(nc_session, rpc, msgid, 1000, &reply);  /* input is freed by parse_reply */

nc_reply_free(reply);
nc_rpc_free(rpc);
lyd_free_withsiblings(input);  /* double free here */
```

If the rpc request data was allocated outside of `parse_reply`, it should not be freed.

Only free it if it was allocated in `parse_reply()`.

### DIFF
--- a/src/session_client.c
+++ b/src/session_client.c
@@ -1232,7 +1232,7 @@ parse_reply(struct ly_ctx *ctx, struct lyxml_elem *xml, struct nc_rpc *rpc, int 
     struct nc_reply_data *data_rpl;
     struct nc_reply *reply = NULL;
     struct nc_rpc_act_generic *rpc_gen;
-    int i;
+    int i, dofree = 1;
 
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
you are right but I thought adding a new flag was not really needed so I fixed it myself. Thanks.

Regards,
Michal

PR #67, opened 2018-08-06
provide correct address length for inet_ntop i.o. to avoid error ENOSPC

frimpler:
Hi Michal,

inet_ntop produces errors with errno "no space left on device" in our environment. According to the man page of inet_ntop this happens when the length of the converted string (which might be up to 45 characters instead of 39 for IPv6 addresses) exceeds the length given to this function.

Regards,
Frank


michalvasko:
Hi Frank,
right, definitely better now, thanks.

Regards,
Michal

PR #69, opened 2018-08-10
Do not send extra ssh_message_reply_default()

### LINKED ISSUE #68: Cannot connect OpenDaylight to netopeer2 with keyboard-interactive authentication
libnetconf2 seems to handle keyboard-interactive authentication incorrectly, as it responds with a spurious SSH_MSG_USERAUTH_FAILURE when asked for keyboard-interactive authentication.
This results in handshake state error, with the an extra SSH_MSG_USERAUTH_FAILURE being delivered to an authenticated session. While OpenSSH reacts with SSH_MSG_UNIMPLEMENTED, Apache MINA SSHD throws an IllegalStateException and terminates the session.

rovarga:
When we are asked to initialize keyboard-interactive authentication,
we should refrain from sending default reply, as that leads to an
extra SSH_MSG_USERAUTH_FAILURE which some clients do not like.

Fixes #68.

Signed-off-by: Robert Varga <robert.varga@pantheon.tech>

### DIFF
--- a/src/session_server_ssh.c
+++ b/src/session_server_ssh.c
@@ -855,6 +855,7 @@ nc_sshcb_auth_kbdint(struct nc_session *session, ssh_message msg)
             char echo[] = {0};
 
             ssh_message_auth_interactive_request(msg, "Interactive SSH Authentication", "Type your password:", 1, prompts, echo);
+            auth_ret = -1;
         } else {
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
you are right it seems and it can be merged. Thanks!

Regards,
Michal

PR #70, opened 2018-08-10
Fix parsing for notification after leaf

alangefe:
This commit fixes parsing of notification messages for the case where the notification follows a leaf. Prior to this change, if the notification followed a leaf, the notification would be missed. In the `LYS_LEAF` case, `elem` would be advanced to `elem->next` (the notification), but the `for` loop condition to detect the notification would not be re-evaluated due to the `goto next_node` jump. By the time the loop condition was re-evaluated, `elem` would have been advanced to the notification's child.

We've written integration tests that demonstrate the problem. This commit, together with https://github.com/ADTRAN/sysrepo/commit/acac046a9421c4ecc0c66d2e33de004bbe92a754, will allow those failing test cases to pass. (See [CI results], lines 7282-7284, or [test source code].) These test cases cover several different variations of notifications tied to data nodes.

[CI results]: https://travis-ci.org/ADTRAN/netopeer2-integration-tests/builds/414144146#L7282
[test source code]: https://github.com/ADTRAN/netopeer2-integration-tests/blob/master/tests/test_notif.py#L233

### DIFF
--- a/src/messages_server.c
+++ b/src/messages_server.c
@@ -16,6 +16,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdarg.h>
+#include <stdbool.h>
 
 #include <libyang/libyang.h>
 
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
nice, thanks!

Regards,
Michal

PR #72, opened 2018-08-15
Use nonzero lock timeout for transmitting messages

alangefe:
This commit adds a nonzero lock timeout for transmitting messages, which greatly improves the reliability of notifications.

The old timeout value of zero meant that if there was any lock contention whatsoever, notifications received from _sysrepo_ that needed to be sent along to clients would be dropped.

We added a nonzero lock timeout to allow the server to wait during the timeout for the lock contention to clear, so that it can transmit the notification.

The reliability improvement is dramatic. Whereas prior to this change, we saw notifications go missing around 3-5% of the time, now we are able to run our tests for more than 1000 iterations without missing a single notification.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -23,6 +23,7 @@
 #include <unistd.h>
 #include <signal.h>
 #include <time.h>
+#include <sys/param.h>
 
 #ifdef NC_ENABLED_TLS
 #   include <openssl/err.h>
@@ -1079,6 +1080,7 @@ nc_write_msg_io(struct nc_session *session, int io_timeout, int type, ...)
     arg.len = 0;
 
     /* SESSION IO LOCK */
+    io_timeout = MAX(io_timeout, NC_IO_LOCK_TIMEOUT);
     ret = nc_session_io_lock(session, io_timeout, __func__);
     if (ret < 0) {
         return NC_MSG_ERROR;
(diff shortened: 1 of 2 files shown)


alangefe:
Hi Michal,

Yes, the problem was not with the send notification calls, but with one of the other IO methods. I don't have the exact details at the moment, so I will follow up on this hopefully next week.

Regards,
Andrew

alangefe:
Hi Michal,

In this case the intermittent notifications were not being sent via `nc_server_notif_send()`, but were being sent via `nc_server_send_reply_io()`. The worker thread was calling `nc_ps_poll()` with a timeout of zero, which would result in `nc_write_msg_io()` being called with `io_timeout = 0`. The method failed to send the notification if there was any IO contention.

The call stack for this scenario is as follows:

```
worker_thread()
nc_ps_poll(timeout=0)
nc_server_send_reply_io(io_timeout=0)
nc_write_msg_io(io_timeout=0)
```

Regards,
Andrew

michalvasko:
Hi Andrew,
maybe I am missing something but I would expect some RPC replies will not be sent in this case because that is where the 0 timeout is used. `nc_server_send_reply_io()` calls `nc_write_msg_io()` only with `NC_MSG_REPLY`. All notifications should be sent (and can block RPC replies) because there is higher timeout used.

The solution would be to simply modify `nc_ps_poll()` timeout but I cannot be sure until the problem is not clear to me.

Regards,
Michal

alangefe:
Hi Michal,

You're correct. On further investigation, we found that the specific changes above aren't actually needed. Sorry for the confusion and thanks for the help.

Regards,
Andrew

PR #73, opened 2018-08-17
Add the yang-library NETCONF capability

alangefe:
This commit adds the yang-library NETCONF capability, which is required by
[RFC 7950 section 5.6.4](https://tools.ietf.org/html/rfc7950#section-5.6.4):

```
    The server also MUST advertise the following capability in the
    <hello> message (line breaks and whitespaces are used for formatting
    reasons only):

      urn:ietf:params:netconf:capability:yang-library:1.0?
        revision=<date>&module-set-id=<id>
```


rkrejci:
Merged manually as 50eac70 with a set of related changes:
- since ietf-yang-library is YANG 1.1 schema, it is not supposed to be part of the capabilities list as a module (only as the capability as changed in this PR)
- the change must reflected when processing capabilities and ietf-yang-library data on a client side to create libyang context according to the server supported schemas.

PR #76, opened 2018-08-27
Provide a mechanism to setup TLS cert chain

alangefe:
This commit from @jwwilcox, together with a corresponding commit to _netopeer2_, fixes the TLS connection scenario in which the server's certificate has been signed by an intermediate CA, but the client only has the root CA available locally. In this case, the client will reject the connection attempt, because it does not know about the intermediate CA.

The changes here use the new _netopeer2_ callback (which supplies the intermediate certificate(s)) to call `SSL_CTX_add_extra_chain_cert()`, which allows the server's TLS context to automatically provide the intermediate certificate(s) to the client.

This scenario is demonstrated in the integration test `test_tls_client_missing_server_intermediate()` in [ADTRAN:netopeer2-integration-tests](https://github.com/ADTRAN/netopeer2-integration-tests/blob/master/tests/test_tls.py#L73). The changes here, together with the corresponding commit in _netopeer2_, will allow [the currently failing test case](https://travis-ci.org/ADTRAN/netopeer2-integration-tests/jobs/420293391#L7434) to pass.

### DIFF
--- a/src/libnetconf.h
+++ b/src/libnetconf.h
@@ -394,7 +394,9 @@
  * So, after starting listening on an endpoint  you need to set the server
  * certificate (nc_server_tls_endpt_set_server_cert()). Its actual content
  * together with the matching private key will be loaded using a callback
- * from nc_server_tls_set_server_cert_clb().
+ * from nc_server_tls_set_server_cert_clb(). Additional certificates needed
+ * for the client to verify the server's certificate chain can be loaded using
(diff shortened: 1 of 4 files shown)


alangefe:
This pull request goes with https://github.com/CESNET/Netopeer2/pull/328.

michalvasko:
Hi,
you are right, it makes sense and was missing before, thanks a lot.

Regards,
Michal

PR #78, opened 2018-09-03
invalid endpoint pointer

pfeige:
Hi,

we are observing sporadically a SIGSEV when netoopeer2-server is performing a call-home. The cause is an invalid endpoint pointer in the client thread of libnetconf2. Within the connect loop the endpoint lock is released and taken again. After such unlock/lock sequence the pointers cur_endpt and cur_endpt_name are in general not more valid because the endpoint arrayof the current client may have changed. Therfore after the lock the mentioned endpoint pointers always have to be determined again. In the original code there are two else branches where this is not done. Here is my proposal for a corresponding fix.

Best Regards,
Peter


michalvasko:
Hi Peter,
thanks, merged. Just a small code formatting issue I will fix.

Regards,
Michal

pfeige:
OK.
Thanks,
Peter

PR #81, opened 2018-09-06
Adapt to CESNET/libyang#602

jktjkt:
API changes for https://github.com/CESNET/libyang/pull/602


michalvasko:
Hi,
seems fine, if updated *libyang* will require it.

Regards,
Michal

PR #82, opened 2018-09-12
Fix call home client thread cleanup

alangefe:
This commit by @thelsper fixes a race condition in call home client creation and deletion. If a call home client is deleted and then immediately re-created with the same client name, the old client thread will not notice that the old client was destroyed, and will attempt to service the new client, while the newly created client thread for the new client also attempts to service the new client! This results in a proliferation of threads and can also result in deadlock.

To fix this, we added a unique client ID to the client information. When a client thread retrieves its client by name, it will now also check the client ID against the cached value of the client ID for the client it was previously servicing. If the client ID changes, this means that the client was destroyed, and the thread will clean itself up.

While we were here making these changes we also changed the session ID to be a true atomic, and removed the spin lock that was protecting it. (stdatomic.h was added in C11, which was why we also appended `-std=gnu11` to the `CFLAGS`. Both sysrepo and netopeer2 are already using C11.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE debug)
 endif()
 
-set(CMAKE_C_FLAGS         "${CMAKE_C_FLAGS} -Wall -Wextra -fvisibility=hidden")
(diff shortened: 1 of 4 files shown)


michalvasko:
Hi,
both changes are fine, thanks, merging.

Regards,
Michal

PR #85, opened 2018-09-20
server session BUGFIX pthread issue on ARM

mislavn:
Fixes issue:
error: 'PTHREAD_RWLOCK_PREFER_WRITER_NONRECURSIVE_NP' undeclared (first use in this function)

I'm not sure what is the correct solution, this patch is based on sysrepo:
https://github.com/sysrepo/sysrepo/blob/master/src/data_manager.c#L2299

Signed-off-by: Mislav Novakovic <mislav.novakovic@sartura.hr>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,8 +172,10 @@ target_link_libraries(netconf2 ${CMAKE_THREAD_LIBS_INIT})
 
 # check availability for some pthread functions
 set(CMAKE_REQUIRED_LIBRARIES pthread)
+include(CheckFunctionExists)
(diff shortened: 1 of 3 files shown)


michalvasko:
Hi Mislav,
in principle it is fine and same mechanism can be used but the availability of this function is checked in CMake so it is missing there. Look at `sysrepo/CMakeLists.txt:143` and `sysrepo/src/common.sr_constants.h.in:33`.

Regards,
Michal

mislavn:
Thanks, I'll implement the changes in the next few days.

Regards,
Mislav

PR #113, opened 2018-11-08
session: Remove engine.h include

neheb:
OpenSSL's engine API is not used except for a cleanup call.


michalvasko:
Hi,
it was there because otherwise there were leaks but I think it was a no longer support OpenSSL version, so fine.

Regards,
Michal

PR #123, opened 2018-12-11
add support for unix socket transport

olivier-matz-6wind:
In addition to SSH and TLS transport, add the support for UNIX
socket. Despite it's not in NETCONF standard, using a UNIX socket has
several advantages:

- no need of password, keys, or certificate. The connection is allowed
  for a local user authenticated on the system, and having the proper
  permissions to connect to the netopeer socket.

- the connection to netopeer is possible even if the firewall is
  enabled. This is an advantage when configuring a firewall equipment
  through netconf.

- the connection to netopeer is possible from another netns.

Signed-off-by: Olivier Matz <olivier.matz@6wind.com>

### DIFF
--- a/python/session.c
+++ b/python/session.c
@@ -324,9 +324,14 @@ newChannel(PyObject *self)
 static PyObject *
 ncSessionStr(ncSessionObject *self)
 {
-    return PyUnicode_FromFormat("NETCONF Session %u to %s:%u (%lu references)", nc_session_get_id(self->session),
-                                nc_session_get_host(self->session), nc_session_get_port(self->session),
-                                ((PyObject*)(self))->ob_refcnt);
+    const char *host = nc_session_get_host(self->session);
+    if (host)
+        return PyUnicode_FromFormat("NETCONF Session %u to %s:%u (%lu references)", nc_session_get_id(self->session),
+                                    nc_session_get_host(self->session), nc_session_get_port(self->session),
+                                    ((PyObject*)(self))->ob_refcnt);
+    else
+        return PyUnicode_FromFormat("NETCONF Session %u to %s (%lu references)", nc_session_get_id(self->session),
+                                    nc_session_get_path(self->session), ((PyObject*)(self))->ob_refcnt);
 }
 
 /*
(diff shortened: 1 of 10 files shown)


michalvasko:
Hi,
thanks for the work, the code seems well written but I am not very keen on merging this into the official repository. All the advantages (so you call them) seem to me like serious breaches of [these](https://tools.ietf.org/html/rfc6241#section-2.2) NETCONF security requirements. Do you really need this to be merged, can't you just use your own patched version?

Regards,
Michal

jktjkt:
From a bystander's perspective, I can think about some use cases for this approach (testing and a local CLI-based access, for example). Also, being able to configure via the NETCONF protocol from a locally running, possibly non-networked access port such as a local console port even in face of a misconfigured firewall looks nice, but then it probably could work when talking via the loopback interface (does your FW support adding rules to loopback? If so, there might be more issues ahead.).

Either way, it seems that a logical next step is implementing `SO_PEERCRED` or `SCM_CREDENTIALS` so that the NETCONF server can infer the Unix identity of the client. That should answer @michalvasko 's reference to NETCONF safety.

olivier-matz-6wind:
Thank you for the feedback.

About security requirements, my understanding is that it they are fulfilled because there is an authentication mechanism:
- the client must be authenticated on the system
- the server can set permissions to the unix socket so that only some uid or gid can connect
- the server gets the netconf username from the unix socket with getsockopt(..., SO_PEERCRED)

As supposed by @jktjkt, we are able use a local cli on our equipment. It can connect to Netopeer with ssh or unix socket. We use a unix socket by default to bypass local firewall rules (our equipment is also a firewall, and it can have a DROP policy by default) and to ensure that the connection can work from another netns.

I'd be happy if it gets merged, and not only because it is easier for us ;) I think it can be useful for other people that have the same use case than us. It guess it is also faster to setup than other transports, for testing purposes.

michalvasko:
I do not doubt the usefulness of this transport, just the security aspects. But I guess it is not as bad as I thought (I am still not sure that communicating using UNIX socket can be considered confidential, but we are talking about the same machine as the server is running on, so fine) so it can be merged.

Regards,
Michal

PR #140, opened 2019-03-13
fix and testcase for empty deviation parameter

frimpler:
In our tests, the \<capability\> element in \<hello\> message sporadically contained an empty deviations parameter ("&deviations="). The issue could be reproduced by a test case (test_server_thread.c).
We see the root cause in the missing initialization of v in in nc_server_get_cpblts_version().

Please have a look,
Frank


michalvasko:
Hi,
that is clearly a bug, thanks for the fix and test.

Regards,
Michal

PR #150, opened 2019-04-23
libssh: Fix the replace order to derive libssh path

chaitu-tk:
In order to derive the library path to include for libssh from the .so
file path, fix the replace order, otherwise you end up with an extra
"/lib" in the path and the build will not be able to find the libssh.


michalvasko:
Hi,
a bit strange that we have never encountered the problem but it seems OK, so it can be merged.

Regards,
Michal

PR #155, opened 2019-06-15
Eliminte two trivial redundancies, which the compiler already optimizes out

adamjrichter:
This pull request eliminates the following two trivial redundancies, both found by cppcheck.

src/messages.c: nc_err_set_msg() assigned the parameter lang to NULL when it already was guaranteed to have that value.  Also, the parameter is never used after that point anyhow.

src/session_server_tls.c: nc_server_tls_get_ctn() had duplicate "&&" clauses ("[...] ctn->name && ctn->name")

These deletions result in unchanged .o files on x86_64 if I build with "-O2" and add a couple of blank lines to messages.c to keep the line numbers the same.

Thanks in advance for considering this patch.

### DIFF
--- a/src/messages_server.c
+++ b/src/messages_server.c
@@ -630,8 +630,6 @@ nc_err_set_msg(struct nc_server_error *err, const char *error_message, const cha
     }
     if (lang) {
         err->message_lang = lydict_insert(server_opts.ctx, lang, 0);
-    } else {
-        lang = NULL;
     }
 
     return 0;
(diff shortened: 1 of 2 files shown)


adamjrichter:
I think the codecov complaints may just be that a redundant line which I deleted happened to be executed by a test, decreasing the covered_lines / all_lines ratio.  I think a better measure would probably be absolute number of uncovered lines, which I believe my patch does not decrease.  However, if I am missing something and there is some change I should make, please just let me know.  Thanks for your attention to this.

michalvasko:
Hi,
no problem, thanks.

Regards,
Michal

PR #160, opened 2019-07-16
Add support for musl libc

neheb:
musl does not support pthread_rwlockattr_setkind_np. Don't use it if it is
not available.

Based on: https://github.com/openwrt/packages/blob/master/libs/libnetconf2/patches/001-fix-pthread-error


michalvasko:
Hi,
okay, merging.

Regards,
Michal

PR #161, opened 2019-07-16
Fix format security warning in newer GCC

neheb:
In OpenWrt, -Werror=format-string is passed, making this fail compilation.

printf style functions are meant to work with format strings anyway.


michalvasko:
Hi,
okay, no problem.

Regards,
Michal

PR #165, opened 2019-08-28
Fix issue where thread specific data is dereferenced inside the key value destructor

kpbarrett:
The destructor nc_client_context_free() calls out to internals which ultimately call nc_client_context_location(). This function uses pthread_getspecific() to get the thread-specific context data, creating it if it does not exist.

When called by the destructor, pthread_getspecific() returns NULL because the thread-specific data is being destroyed. (see pthread_key_create(3)). This causes a thread-specific context to be allocated while the thread's context is being destroyed, leading to nasty effects including segfaults especially when more than one thread is used.

The change replaces the internal API calls with calls which use a specific context location.

### DIFF
--- a/src/session_client.c
+++ b/src/session_client.c
@@ -97,15 +97,25 @@ nc_client_context_free(void *ptr)
 #endif
     {
         /* for the main thread the same is done in nc_client_destroy() */
-        nc_client_set_schema_searchpath(NULL);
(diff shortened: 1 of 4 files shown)


michalvasko:
Hi,
seems correct, thanks.

Regards,
Michal

PR #167, opened 2019-09-08
Modify for avoid deprecated functions of libssh

brchiu:
ssh_get_publickey() becomes deprecated function from libssh 0.8.0.

ssh_get_publickey() and ssh_write_knownhost() become deprecated
functions from libssh 0.9.0.


michalvasko:
Hi,
no problem, merged.

Regards,
Michal

PR #169, opened 2019-09-09
Wrong umask prevents temporary private key access

mekleo:
Hi,
umask 0600 prevents the user from reading his own temporary private key file.
0177 might be the intended mask.
Kind regards.


michalvasko:
Hi,
you are right, thanks.

Regards,
Michal

PR #175, opened 2019-10-15
io fixes on non-blocking FDs

rjarry:
These are fixes for problems I encountered with the UNIX transport. Basically, when sending large configurations to netopeer2-server, since the socket fd is non-blocking, we can get `EAGAIN` errors upon `write()` if the socket buffer is full and the server does not read fast enough.

These errors are already handled in libnetconf2 but errno is not reset and libyang checks it (since commit https://github.com/CESNET/libyang/commit/609d708ca567d5400f303e654e28637203e1d07e). This causes the operation to fail with the following error:

```
Print error (Resource temporarily unavailable).
```

Work around the problem by resetting `errno` to `0` in case of `EAGAIN` or `EINTR`.

Also, if the error is `EINTR` do not wait before trying to read/write again.

I am not sure this is the correct approach, I proposed another (maybe complementary?) solution in https://github.com/CESNET/libyang/pull/903

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -75,7 +75,7 @@ nc_read(struct nc_session *session, char *buf, size_t count, uint32_t inact_time
 {
     size_t readd = 0;
     ssize_t r = -1;
-    int fd;
+    int fd, interrupted;
     struct timespec ts_cur, ts_inact_timeout;
 
     assert(session);
@@ -92,6 +92,7 @@ nc_read(struct nc_session *session, char *buf, size_t count, uint32_t inact_time
     nc_gettimespec_mono(&ts_inact_timeout);
     nc_addtimespec(&ts_inact_timeout, inact_timeout);
     do {
+        interrupted = 0;
         switch (session->ti_type) {
         case NC_TI_NONE:
             return 0;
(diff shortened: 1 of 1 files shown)


brchiu:
@rjarry  
Thanks for addressing this issue. I also encountered a similar one for SSH case with `Print error (No such file or directory)` error message. Can your approach apply to it ?

src/io.c
```diff
@@ -190,7 +197,8 @@ nc_read(struct nc_session *session, char *buf, size_t count, uint32_t inact_time
 
         if (r == 0) {
             /* nothing read */
-            usleep(NC_TIMEOUT_STEP);
+            if (!interrupted)
+                usleep(NC_TIMEOUT_STEP);
```
michalvasko: Please always use `{` and `}`.

michalvasko:
Hi,
depending on how we resolve cesnet/libyang#903, resetting `errno` may not be needed.

Regards,
Michal

rjarry:
> @rjarry
> Thanks for addressing this issue. I also encountered a similar one for SSH case with `Print error (No such file or directory)` error message. Can your approach apply to it ?

@brchiu I don't think so, I only handle `EAGAIN` and `EINTR`, `No such file or directory` is `ENOENT` and is a real error.

src/io.c
rjarry: done.

rjarry:
> depending on how we resolve [CESNET/libyang#903](https://github.com/CESNET/libyang/pull/903), resetting `errno` may not be needed.

Indeed, I removed the first commit.

michalvasko:
Hi,
it is fine now, thanks.

Regards,
Michal

rjarry:
:+1:

PR #177, opened 2019-10-16
session_client: fix segfault on NC_MSG_REPLY_ERR_MSGID

rjarry:
When `retrieve_schema_data_getschema` or `build_schema_info_yl` receive an `NC_MSG_REPLY_ERR_MSGID` error, they do not call `nc_recv_reply` again to receive the expected message.

However, when `msg == NC_MSG_REPLY_ERR_MSGID`, `reply` is necessarily `NULL`, which causes a segfault when accessing the `reply->type` field.

Call `nc_recv_reply` again if we receive an `NC_MSG_REPLY_ERR_MSGID` error.

Also, `nc_recv_reply` may leave `reply` `NULL` for other reasons (`NC_MSG_ERROR`
is not the only case). Return an error if reply is `NULL` to avoid potential further segfaults.

### DIFF
--- a/src/session_client.c
+++ b/src/session_client.c
@@ -379,12 +379,12 @@ retrieve_schema_data_getschema(const char *name, const char *rev, struct clb_dat
 
     do {
         msg = nc_recv_reply(clb_data->session, rpc, msgid, NC_READ_ACT_TIMEOUT * 1000, 0, &reply);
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
okay, no problem, thanks.

Regards,
Michal

PR #178, opened 2019-10-16
session_client: fix user searchpath for unix/fd transports

rjarry:
The `nc_session_new_ctx()` function creates a fresh libyang context (unless an explicit context has been provided) and initializes the context search dirs with the user search path (if set) and `NC_SCHEMAS_DIR`.

This function is only called for the SSH and TLS transports. Making the user searchpath ignored when using the UNIX and FD transports. Call it for all transports.

Fixes: fd5b66899d66 ("client CHANGE rewrite loading the schemas into context when connecting to a server")
Fixes: ac7fa2fb48a3 ("add support for unix socket transport")

### DIFF
--- a/src/session_client.c
+++ b/src/session_client.c
@@ -1098,18 +1098,10 @@ nc_connect_inout(int fdin, int fdout, struct ly_ctx *ctx)
     session->ti.fd.in = fdin;
     session->ti.fd.out = fdout;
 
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
seems fine, thanks.

Regards,
Michal

rjarry:
Thanks @michalvasko !

PR #197, opened 2020-01-09
Export cmake config files

ClausKlein:
cmakelint all cmake files too
fix errors in cmake modules used.
prevent compile errors if libcrypt is not available on BSD/OSX

see too https://github.com/Bagira80/More-Modern-CMake

NOTE: at least cmake v3.9 is required!

The cmake config files are exported now to be used by other cmake projects  with find_package():

-- Installing:
/usr/local/lib/cmake/libnetconf2/libnetconf2-config-version.cmake
-- Installing: /usr/local/lib/cmake/libnetconf2/libnetconf2-targets.cmake
-- Installing:
/usr/local/lib/cmake/libnetconf2/libnetconf2-targets-release.cmake
-- Installing: /usr/local/lib/cmake/libnetconf2/libnetconf2-config.cmake


ClausKlein:
Please check the files under packages, I had problems with git rebase devel; sorry

michalvasko:
Hi,
okay, I have looked a bit more into the currently supported distributions and it is not looking good. So, unfortunately, I must require that the highest required `cmake` version to be only `2.8.12`. Also, I am not a fan of including some optional statements in case the used `cmake` is of a higher version and would prefer to add these features once the particular `cmake` version is strictly required. So, if you can stick to this (which I think would still keep at least half this pull request, which are only fixes and not changes), I will merge it.

Also, you have included (maybe by a mistake) some reverts of changes I have made before (such as using ECDSA keys in tests instead of DSA) so please keep those.

Regards,
Michal

ClausKlein:
Hi,

I understand. I may split the patches ones again. But not this month, sorry, I have no more time.

About cmake v2.8.12: There are to many fixes and improvements at current cmake; and this is v3.16!

I am not willing to work with cmake version less than v3.9.

With Regards,
Caus

„Es ist nicht genug, zu wissen, man muß auch anwenden; 
     es ist nicht genug, zu wollen, man muß auch tun.“ 
Johann Wolfgang von Goethe <http://www.zitate-online.de/autor/goethe-johann-wolfgang-von/> (Werk: Wilhelm Meisters Wanderjahre)

michalvasko:
Hi,

> I understand. I may split the patches ones again. But not this month, sorry, I have no more time.
No problem, that is up to you.

> I am not willing to work with cmake version less than v3.9.
I understand. Then it will be up to us to gradually improve things and finally move to some newer `cmake` version once it will be possible.

Regards,
Michal

michalvasko:
I forgot to mention that you do not have to worry about the packaging commands and corresponding files, I will clean and refactor that a bit myself soon.

Regards,
Michal

PR #200, opened 2020-01-13
python: do not put generated files into the source tree

jktjkt:
Untracked content shows up in git submodules, causing stuff to be marked "dirty" by various git-level tools.

Tested by temporarily putting the resulting target within `ALL`. The result is the same as without this change (and it remains unusable with no generated index, at least when building out-of-tree, BTW).


michalvasko:
Hi,
okay, no problem.

Regards,
Michal

PR #210, opened 2020-02-12
URL capability MUST contain a scheme

mekleo:
According to https://tools.ietf.org/html/rfc6241#section-8.8.3, the URL Capability Identifier MUST contain a scheme argument.

Hence, this capability cannot be merely infered from ietf-netconf's features. It has to be explicitly provided using for example
nc_server_set_capability("urn:ietf:params:netconf:capability:url:1.0?scheme=scp,http,https,ftp,sftp,ftps,file");

Since add_cpblt() cannot overwrite capabilities, the incomplete one "urn:ietf:params:netconf:capability:url:1.0" should not be added in the first place.

This bug prevents yangcli from proposing url leaves, as no protocol can be validated against that incomplete capability identifier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.6)
 
-project(libnetconf2 LANGUAGES C)
+project(libnetconf2 C)
 
 include(GNUInstallDirs)
 include(CheckFunctionExists)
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi,
you are right, both the incorrect and correct capabilities are set but only the first wrong one is sent. Your fix is fine but please change the branch base to `devel` and just comment the code out with a note explaining that the capability must be set manually because of the need for supported protocols to be included. Thanks.

Regards,
Michal

michalvasko:
Much better, thanks, merging.

Regards,
Michal

PR #214, opened 2020-03-24
Error log fixed

ravsz:
Currently error logs looks like: 'Call Home client "(null)" is not of periodic connection type.'


michalvasko:
Right, thanks.

Regards,
Michal

PR #220, opened 2020-04-14
libnetconf2 CHANGE QNX compatibility

### LINKED ISSUE #217: libnetconf2 support on QNX platform
Hi,

I'd like to port libnetconf2 over to QNX. Below I've summarized the issues in no particular order.
Let me know if you want more details for any of these issues. Aside from issue 2, I can submit patches for the rest. If you do want patches, how would you want them divided?

1. `fd_set` and friends are defined in `#include <sys/select.h>` on QNX. The relevant file is `session_client.c`.
2. `asprintf`, `vasprintf` and `strndup` are used in `io.c`, `messages_server.c`, `session.c` and `session_client.c` but  are not available on QNX. Would it be possible to provide a `compat.c` similar to this [commit](https://github.com/CESNET/libyang/commit/0c5ec1bf4a0c7c49afe4f8b32951b105130c8177) in libyang?
3. `#include <sys/syscall.h>` is not available on QNX. It's used in `session_client.c`. Would it be possible to put it in a `#ifdef __linux__` wrapper, as the only time it's called is in a section of code also delimited by `#ifdef __linux__`?
4. In `session_server.c`, `nc_accept_unix` relies on `SO_PEERCRED`. I think it would be possible to provide an alternate implementation that would cover the BSDs and as a result QNX (networking stack is NetBSD) and perhaps MAC (due to Darwin). `getpeereid` could be used to get the `euid` which in turn could be used for the rest of `nc_accept_unix`. See [here](https://github.com/dun/munge/wiki/Auth-Methods) for details and [here](http://www.qnx.com/developers/docs/6.5.0/index.jsp?topic=%2Fcom.qnx.doc.neutrino_lib_ref%2Fg%2Fgetpeereid.html) for QNX documentation. What are your thoughts on this?
5. QNX links against `-llogin` instead of `-lcrypt` for the `crypt()` function. See `target_link_libraries(netconf2 "-L${LIBSSH_LIBRARY_DIR}" -lssh -lcrypt)` in `CMakeLists.txt`.
6. In `session_server_ssh.c`, QNX doesn't provide `#include <crypt.h>`. It does however provide `#include <shadow.h>`.
7. In `session_server_ssh.c`, `getspnam_r` is used in `auth_password_get_pwd_hash`. `getspnam_r` is defined differently for QNX compared to linux. In QNX it returns what would otherwise be obtained from the fifth parameter. Here's the [doc](www.qnx.com/developers/docs/6.5.0/index.jsp?topic=%2Fcom.qnx.doc.neutrino_lib_ref%2Fg%2Fgetspnam.html).

apropp-molex:
fixes #217 
Below is a summary of the compatibility issues this solves:
1. Links against `liblogin` for QNX only
1. Checks if `crypt.h`, `shadow.h` exists instead of relying on OS checks
1. Accommodates different function signature for `getspnam_r` under QNX
1. Uses `getpeereid` for QNX only. This is because `-D_XOPEN_SOURCE=500` in CMakeLists.txt caused conflict with `-D_POSIX_C_SOURCE=200809L` and using `#define _XOPEN_SOURCE 500` in `session_server.c` caused a host of issues with pthread stuff. `libpthread` is removed from CMAKE_REQUIRED_LIBRARIES as it cmake failed to link while looking for `getpeereid`. `libpthread` is replaced by `libc` on QNX.
Other than those issues, I don't have a BSD/MAC system to test this against, so restricted it to QNX to avoid causing regression issues.

### DIFF
--- a/src/session_server.c
+++ b/src/session_server.c
@@ -11,6 +11,7 @@
  *
  *     https://opensource.org/licenses/BSD-3-Clause
  */
+#define _QNX_SOURCE /* getpeereid */
 #define _GNU_SOURCE /* signals, threads, SO_PEERCRED */
 
 #include <stdint.h>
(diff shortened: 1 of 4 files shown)


src/session_server.c
```diff
@@ -1720,24 +1721,35 @@ nc_ps_clear(struct nc_pollsession *ps, int all, void (*data_free)(void *))
...
+        }
+        uid = ucred.uid;
+    #else
+        if (getpeereid(sock, &uid, NULL) < 0) {
+            ERR("Failed to get credentials from unix socket (%s).",
+                strerror(errno));
+            close(sock);
+            return -1;
+        }
+    #endif
```
michalvasko, line 1748: I will ask you to move this into a function so that `nc_accept_unix` does not have any `#ifdef`s, thank you.

apropp-molex: Moved into function called `nc_get_uid`. Let me know if that's what you were expecting.

michalvasko:
Hi,
almost, I will add some final minor improvements myself. Merging.

Regards,
Michal

PR #223, opened 2020-04-20
build CHANGE remove safety check

apropp-molex:
Refs #217
Removing the check for `libcrypt`. One reason is that `libcrypt` is linked later on so any checking should be done there. Second reason is some platforms might implement this functionality in a different library (e.g., QNX with `liblogin`). Also, the check can be bypassed by setting `-DLIBCRYPT`.
The failure will be clear enough to indicate to the user what went wrong during compilation. Opted not to use `find_library` to look for the appropriate library, as couldn't get it to find `liblogin`.


michalvasko:
Okay, merging.

Regards,
Michal

PR #229, opened 2020-05-15
io BUGFIX do not fail when there is data left to read

rjarry:
When using the UNIX transport and depending on the timing between the server and client, the client may read the final reply to the `<close-session>` rpc __after__ the server has closed the socket. When that occurs, `poll()` returns `POLLHUP` in `revents` meaning the server closed the socket which causes the following error on the client:

```
ERR: Session 1: communication channel unexpectedly closed.
ERR: Session 1: failed to receive a reply to <close-session>.
```

When `POLLHUP` is returned in `revents`, make sure `POLLIN` is not present as well (if it is, it means that there is some data left to be read in the socket buffer).

See https://www.greenend.org.uk/rjk/tech/poll.html for more details.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -605,16 +605,18 @@ nc_read_poll(struct nc_session *session, int io_timeout)
         return -1;
     } else { /* status > 0 */
         /* in case of standard (non-libssh) poll, there still can be an error */
-        if (fds.revents & POLLHUP) {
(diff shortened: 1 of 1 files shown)


michalvasko:
Seems good to me and it makes sense, thanks, merging.

Regards,
Michal

rjarry:
Thanks Michal :+1:

PR #233, opened 2020-05-19
Python BUGFIX make install fail when running from path with a space

jakubman1:


michalvasko:
Right, thanks.

Regards,
Michal

PR #236, opened 2020-05-28
Get the alarm content when connect_ssh_session is failed

eshang-oran:
As netconf client, we want to get the alarm content when connect_ssh_session is failed.
We can get the alarm content by call libssh interface.


michalvasko:
Hi,
firstly, please always target `devel` branch, we will not merge anything directly into `master`. Secondly, you added the SSH error print into a general fail, which is most often not related to SSH at all. So, please point me the *libssh* function that can fail and whose error is not handled and then something can be improved (I have not found any such call in this function and any called functions).

Regards,
Michal

PR #254, opened 2020-09-10
Add flag to ignore libssh version

syyyr:
When building with a new enough libssh version that the issues are
fixed, libnetconf2 still considers these broken. This patch adds an
option to skip this check and use whatever version is available.

For example, Fedora already has patches for the issues that were present in the blacklisted versions.


michalvasko:
Is this really necessary? Firstly, *libssh* compiled from `master` should itentify itself as 0.8.90, to my knowledge, unless something got changed. As for some patched versions presenting themselves as 0.9.4, that is where the problem is apparent but I am not sure it is worth cluttering `CMakeLists` with another option. You can bypass this by, for example, changing the version in the header to 0.9.5 or something (which seems correct to me in some way, it is not 0.9.4). Anyway, I am hoping that another *libssh* release will come soon and then this problem will be less pressing.

syyyr:
Yeah... I guess you're right. I asked the libssh devs on IRC, the 0.9.5 release should come very soon. Until then I'll just use this patch as it is, without merging into libnetconf2.

michalvasko:
Okay, good.

PR #272, opened 2020-12-10
Update session.c

Rohit2387:
NULL check


michalvasko:
This is fixed in `devel`, please use that until the next release.

Rohit2387:
OK thanks then i can fork devel?

michalvasko:
You can fork whatever branch you want. But you should always use either `master` or `devel` branches (of *libyang*, *sysrepo*, *libnetconf2*, and *netopeer2*) to make sure it is compatible.

PR #276, opened 2020-12-16
Sample App usage for libnetconf2 ssh server

Rohit2387:
Hi,

I have created one sample file , so that some one who are new to libnetconf2, can look this sample code and understand the API usage part.

Br
Rohit

### DIFF
--- a/sample_ssh_netconf_server_api_usage.c
+++ b/sample_ssh_netconf_server_api_usage.c
@@ -1,0 +1,129 @@
+#include <pthread.h>
+#include <stdlib.h>
+#include <stdio.h>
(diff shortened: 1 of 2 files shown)


michalvasko:
Thanks for the effort, I am sure you learnt some basics by working on this, but it would need major improvements before it could be merged. So I think you can just use it yourself.

Rohit2387:
Ok thanks for your comment.
This sample code is very basic according to you  looking this sample code is API usage is proper?

michalvasko:
Yes, the usage of API functions seems correct to me.

PR #279, opened 2020-12-28
Add some testcases

fredgan:
Hi Michal,
I add some testcases to `test_fd_comm.c` and `test_io.c`.
In addition, `ietf-netconf-nmda.yin` and its dependencies can not be found in the `tests/data/module` path, so I added them.
https://github.com/CESNET/libnetconf2/blob/819598bdfafb09f9a8bb49e57c2d6d0aac5208b3/src/session_client.c#L2363-L2369


michalvasko:
After looking at the current tests, I would say they are not that good in general and your additions are making things even worse (especially `test_io` where you are overwriting work done in `setup`). So, thanks for the effort, but I am reluctant to merge this and I would either prefer to keep the limited tests as they are or properly update and fix them. I will most likely get to it when I will be migrating *libnetconf2* to *libyang* version 2.

PR #280, opened 2020-12-30
add some  testcases for client call home

fredgan:


michalvasko:
Okay, I suppose, these tests seem fine. Thanks.

PR #283, opened 2021-01-13
Fix libyang error when setting confirmed commit

syyyr:
When creating a commit RPC, if the ietf-netconf module doesn't have the
"confirmed-commit" feature enabled, it always prints a libyang error. This
patch adds a check for that.

Reproducer program:
```cpp
#include <libnetconf2/session_client.h>

int main(int argc, char* argv[])
{
    auto sess = nc_connect_unix("/opt/cesnet/Netopeer2/netopeer2-server.sock", nullptr);
    auto rpc = nc_rpc_commit(1, 0, nullptr, nullptr, NC_PARAMTYPE_CONST);
    uint64_t id;
    nc_send_rpc(sess, rpc, 0, &id);
}
```
When run, it prints:
```
$ ./a.out 
libyang[0]: Failed to find "confirmed" as a sibling to "ietf-netconf:confirmed".
```

### DIFF
--- a/src/session_client.c
+++ b/src/session_client.c
@@ -2661,13 +2661,15 @@ nc_send_rpc(struct nc_session *session, struct nc_rpc *rpc, int timeout, uint64_
         rpc_com = (struct nc_rpc_commit *)rpc;
 
         data = lyd_new(NULL, mod, "commit");
-        if (rpc_com->confirmed) {
-            lyd_new_leaf(data, mod, "confirmed", NULL);
-        }
+        if (lys_features_state(mod, "confirmed-commit")) {
+            if (rpc_com->confirmed) {
+                lyd_new_leaf(data, mod, "confirmed", NULL);
+            }
 
-        if (rpc_com->confirm_timeout) {
-            sprintf(str, "%u", rpc_com->confirm_timeout);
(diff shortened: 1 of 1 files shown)


michalvasko:
So you want to silently ignore these parameters? I definitely consider printing an error a better solution, it is intentional.

syyyr:
Yeah, you are right, this isn't a solution, my bad. So, what if, when the server doesn't support confirmed-commit, the RPC fails to send and libnetconf2 prints a custom error message, something like "Confirmed commit is enabled on the remote server"? When I first used nc_rpc_commit I didn't realize what "confirmed" means, so I just put in "1" (which is my mistake, I admit). Right now, if you put in "1" and the server doesn't support it, libnetconf2 still ignores it and prints a libyang error, but, to be honest, the libyang error doesn't help too much...

michalvasko:
Right, the return value is not checked for these leaves but it is hard to decide whether it was intentional or not. In any case, I have added the checks. As for the error, I agree it is not really clear, but there is not much to be done about it because it is specified this way (although I am not able to find this right now). Also, *libyang* version 2 is not even able to determine whether the node does not exist or has only false `if-feature` so it will not be changed.

PR #295, opened 2021-03-09
Fix _nc_connect_libssh port on big endian processors

dragosda:
See, gets cast to (unsigned int*):
1561 ssh_options_get_port(ssh_session, (unsigned int *)&port);


dragosda:
See, gets cast to (unsigned int*):
1561     ssh_options_get_port(ssh_session, (unsigned int *)&port);

michalvasko:
Right, definitely a wrong code, merging.

PR #303, opened 2021-05-21
keep track of NC-sessions-SR-session associations

frimpler:
see https://github.com/CESNET/netopeer2/issues/901


michalvasko (changes requested):
I must ask for one change, at least. Make the refcount `ATOMIC_UINT32_T` (from `config.h`) to prevent thread sanitizers from complaining. You will need to add `ATOMIC_DEC` macro, but that should be trivial.

PR #304, opened 2021-05-24
build CHANGE separated build type for building docs without dependencies

rkrejci:
Let the user build doxygen documentation without need of other compile dependencies.


rkrejci:
same as #305 just for a different target branch

michalvasko (commented):
I am thoroughly confused as to what is going on but these changes are already present in [libyang2](https://github.com/CESNET/libyang/blob/libyang2/CMakeLists.txt#L24).

michalvasko (approved):

PR #319, opened 2021-07-14
distro FEATURE upstream RPM packaging

jruzicka-nic:
Initial RPM packaging - might miss some BuildRequires (like `libssh-devel` ?) but it builds in my Fedora 34 VM against master `libyang2` package.

Good enough to start some packaging CI a la `libyang2` I reckon.

In related news, latest `apkg-0.2.0` features improved `apkg install` which can be used without parameters to build and install packages from project source in one step - especially useful in containers.


distro/pkg/rpm/libnetconf2.spec
```diff
@@ -0,0 +1,64 @@
...
+Release: {{ release }}%{?dist}
+Summary: NETCONF protocol library
+Url: https://github.com/CESNET/libnetconf2
+Source: libnetconf2-%{version}.tar.gz
+License: BSD-3-Clause
+
+BuildRequires:  cmake
+BuildRequires:  gcc
+BuildRequires:  make
+BuildRequires:  libyang2-devel
```
michalvasko, line 12: As you noticed, `libssh-devel` but also `libopenssl-devel` should be added as `BuildRequires`. I am also confused by any `Requires` missing (such as `libyang2`, `libssh`, and `openssl`), that should not be included?

michalvasko (changes requested):
Only minor issues, seems fine generally.

distro/pkg/rpm/libnetconf2.spec
jruzicka-nic: I've added the missing `BuildRequires`.

For C/C++ libs, modern `rpm` is able to detect runtime `Requires` quite well so there is no need to specify those explicitly AFAIK.

Here is a relevant portion of `rpm -qp --requires libnetconf2-2.0.1.2.gbe78a4c-1.fc34.x86_64.rpm` output:

```
libcrypt.so.2()(64bit)
libcrypt.so.2(XCRYPT_2.0)(64bit)
libcrypto.so.1.1()(64bit)
libcrypto.so.1.1(OPENSSL_1_1_0)(64bit)
libpthread.so.0()(64bit)
libpthread.so.0(GLIBC_2.2.5)(64bit)
libpthread.so.0(GLIBC_2.3.2)(64bit)
libssh.so.4()(64bit)
libssh.so.4(LIBSSH_4_5_0)(64bit)
libssl.so.1.1()(64bit)
libssl.so.1.1(OPENSSL_1_1_0)(64bit)
libyang.so.2()(64bit)
```

michalvasko: Alright, sounds good, did not know that.

michalvasko (approved):

michalvasko:
Should be okay now, is there anything else needed (WiP?) or can be merged? If you want, you can add building this package into `.github/workflows/ci.yaml` as a new matrix config `Package`.

jruzicka-nic:
> Should be okay now, is there anything else needed (WiP?) or can be merged? If you want, you can add building this package into `.github/workflows/ci.yaml` as a new matrix config `Package`.

As this is a new feature that has no chance of breaking existing functionality, I'd merge it.

I'm not overly familiar with GitHub CI but I'll give the `Package` CI a shot and see what I end up with - but that can be done in a dedicated MR.

Biggest problem there is getting the `libyang2` package installed... I guess that could be done during image setup? We'll see ;)

michalvasko:
> Biggest problem there is getting the libyang2 package installed...

Right, in the current CI "targets" it is compiled and installed from source, which will not do. And if it is meant to work for `devel` I think the only possibility is to build the latest package of *libyang2* manually and then install it, otherwise you will almost never get the latest one. What about installing the dependencies (only *libyang2* in this case) from source without a package like it is now and force `apkg` to ignore the specific package dependency? It should test building the project package reliably but not require building any other packages.

jruzicka-nic:
> What about installing the dependencies (only _libyang2_ in this case) from source without a package like it is now and force `apkg` to ignore the specific package dependency? It should test building the project package reliably but not require building any other packages.

TBH it would be less problematic to simply clone `libyang2` source and run `apkg install` on that to build and install `libyang2` packages.

Also depends whether you want to tests against latest available or latest released/packaged `libyang2`. Simply waiting until `libyang2` makes it into Fedora ([review in progress](https://bugzilla.redhat.com/show_bug.cgi?id=1982306)) and using that might be a solution too... but that doesn't really cover all platforms and all use cases.

PR #348, opened 2021-11-25
libnetconf2 BUGFIX netopeer2 build fails when building with musl libc

hthiery:
while updating buildroot package this failure was observed during package testing


michalvasko:
Fixed myself, `uncrustify` complains if the headers are not in alphabetical order.

PR #352, opened 2021-12-12
WIP: Sync the Debian packaging with in-Debian packaging

oerdnj:
The commit message and PR description needs to be updated first.


michalvasko:
So I can change the commit message when merging but I do not know what to write there, why is the compat file no longer needed?

oerdnj:
> So I can change the commit message when merging but I do not know what to write there

I'll do that, no worries.

> why is the compat file no longer needed?

The first commit sync the current packaging work back-and-forth (I need to tweak `sysrepo` -> `sysrepo-tools` and various other little tweaks in current Debian packages) and there will be more commits on top of this. One of them would be to change the compatibility back to <foo>. Where <foo> at the moment is I guess Debian stretch (EOL June 2022) and Ubuntu 18.04 (EOL April 2023). Is there anything else that you absolutely must support?

michalvasko:
No, what you mentioned is as old as we would like to support, thanks.

PR #363, opened 2022-02-03
Update libnetconf2-dev.install

wungad:
This file has to be present for successful CESNET/netopeer2 build (devel).
https://github.com/CESNET/netopeer2/blob/devel/CMakeModules/FindLibNETCONF2.cmake#L74


michalvasko:
Yes, thanks.

PR #370, opened 2022-04-20
rpm UPDATE packaging from Fedora

jruzicka-nic:
Fedora package sources:

https://src.fedoraproject.org/rpms/libnetconf2

### DIFF
--- a/distro/pkg/rpm/libnetconf2.spec
+++ b/distro/pkg/rpm/libnetconf2.spec
@@ -3,11 +3,10 @@ Version: {{ version }}
 Release: {{ release }}%{?dist}
 Summary: NETCONF protocol library
 Url: https://github.com/CESNET/libnetconf2
-Source: libnetconf2-%{version}.tar.gz
+Source: %{url}/archive/v%{version}/%{name}-%{version}.tar.gz
 License: BSD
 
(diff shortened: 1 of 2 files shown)


distro/pkg/rpm/libnetconf2.spec
```diff
@@ -28,27 +28,22 @@ servers. NETCONF is the NETwork CONFiguration protocol introduced by IETF.
...
 %build
-cd build
-cmake \
-    -DCMAKE_INSTALL_PREFIX:PATH=%{_prefix} \
-    -DCMAKE_BUILD_TYPE:String="Release" \
-    -DCMAKE_C_FLAGS="${RPM_OPT_FLAGS}" \
-    -DCMAKE_CXX_FLAGS="${RPM_OPT_FLAGS}" \
-    ..
-make
+%cmake
```
michalvasko, line 33: `-DCMAKE_BUILD_TYPE=RELWITHDEBINFO` is not needed, does the macro supply any specific build type? Maybe just for consistency it should be added anyway, was required for the *sysrepo* package.

michalvasko (changes requested):

michalvasko (approved):

PR #379, opened 2022-06-20
kbdint update auth using Linux PAM

Roytak:
new way to authenticate, it's test, and merged time functions


CMakeModules/FindLibPAM.cmake
```diff
@@ -0,0 +1,82 @@
...
+  find_path(LIBPAM_INCLUDE_DIR
+    NAMES
+      security/pam_appl.h
+    PATHS
+      /usr/include
+      /usr/local/include
+      /opt/local/include
+      /sw/include
+      ${CMAKE_INCLUDE_PATH}
+      ${CMAKE_INSTALL_PREFIX}/include
```
jktjkt, line 49: CMake searches in several of these paths by default, and some others (`/sw`, for example) are very non-standard. Let's stick to the defaults here.

CMakeModules/FindLibPAM.cmake
```diff
@@ -0,0 +1,82 @@
...
+      pam
+    PATHS
+      /usr/lib
+      /usr/lib64
+      /usr/local/lib
+      /usr/local/lib64
+      /opt/local/lib
+      /sw/lib
+      ${CMAKE_LIBRARY_PATH}
+      ${CMAKE_INSTALL_PREFIX}/lib
```
jktjkt, line 63: same here

src/io.c
```diff
@@ -211,9 +210,8 @@ nc_read(struct nc_session *session, char *buf, size_t count, uint32_t inact_time
             if (!interrupted) {
                 usleep(NC_TIMEOUT_STEP);
             }
-            nc_gettimespec_mono(&ts_cur);
-            if ((nc_difftimespec(&ts_cur, &ts_inact_timeout) < 1) || (nc_difftimespec(&ts_cur, ts_act_timeout) < 1)) {
-                if (nc_difftimespec(&ts_cur, &ts_inact_timeout) < 1) {
+            if ((nc_difftimespec_cur(&ts_inact_timeout) < 1) || (nc_difftimespec_cur(ts_act_timeout) < 1)) {
+                if (nc_difftimespec_cur(&ts_inact_timeout) < 1) {
```
jktjkt, line 214: How is this change related to adding support for PAM? If a timer rework is required, let's do that in a separate commit, please.

jktjkt (changes requested):
This PR contains two new features, let's use one PR for adding the examples, and one PR for adding the PAM integration. Do not add `*.so` to git, and please avoid merge commits when submitting a PR -- just rebase your changes on top of most recent `devel`.

src/io.c
Roytak: My apologies, this was not supposed to be an actual pull request, hence the .so. Anyways, I appreciate your advice.

PR #382, opened 2022-06-30
libnetconf2 FEATURE auth using Linux PAM

Roytak:
Added the ability to authenticate via Linux PAM when using keyboard-interactive SSH authentication method. With Linux PAM (or Pluggable Authentication Modules for Linux) you are able to create your own type of authentication service by creating a PAM module (a test module is a part of this PR). One new API call was added. Also two new timespec functions were added by merging old ones in order to make some conditions easier to read and less LoC overall.

### DIFF
--- a/CMakeModules/FindLibPAM.cmake
+++ b/CMakeModules/FindLibPAM.cmake
@@ -1,0 +1,87 @@
+# - Try to find LibPAM
+# Once done this will define
+#
+#  LIBPAM_FOUND - system has LibPAM
+#  LIBPAM_INCLUDE_DIRS - the LibPAM include directory
+#  LIBPAM_LIBRARIES - link these to use LibPAM
+#  LIBPAM_HAVE_CONFDIR - LibPAM version >= 1.4
+#
+#  Author Roman Janota <xjanot04@fit.vutbr.cz>
+#  Copyright (c) 2022 CESNET, z.s.p.o.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#  1. Redistributions of source code must retain the copyright
+#     notice, this list of conditions and the following disclaimer.
+#  2. Redistributions in binary form must reproduce the copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#  3. The name of the author may not be used to endorse or promote products
+#     derived from this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+#  IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+#  OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+#  IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT,
+#  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
+#  NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+#  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+#  THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+#  THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+if(LIBPAM_LIBRARIES AND LIBPAM_INCLUDE_DIRS)
+  # in cache already
+  set(LIBPAM_FOUND TRUE)
+else()
+
+  find_path(LIBPAM_INCLUDE_DIR
+    NAMES
+      security/pam_appl.h
+      security/pam_modules.h
+    PATHS
+      /opt/local/include
+      /sw/include
+      ${CMAKE_INCLUDE_PATH}
+      ${CMAKE_INSTALL_PREFIX}/include
+  )
+
+  find_library(LIBPAM_LIBRARY
+    NAMES
+      pam
+    PATHS
+      /usr/lib
+      /usr/lib64
+      /opt/local/lib
+      /sw/lib
+      ${CMAKE_LIBRARY_PATH}
+      ${CMAKE_INSTALL_PREFIX}/lib
+  )
+
+  if(LIBPAM_INCLUDE_DIR AND LIBPAM_LIBRARY)
+    set(LIBPAM_FOUND TRUE)
+    # there is no reliable way to check the version of PAM, so see if the declaration of a function pam_start_confdir
+    # is in the header file (added in PAM 1.4)
+    file(STRINGS ${LIBPAM_INCLUDE_DIR}/security/pam_appl.h PAM_CONFDIR REGEX "pam_start_confdir")
+    if ("${PAM_CONFDIR}" STREQUAL "")
+      set(LIBPAM_HAVE_CONFDIR FALSE)
+    else()
+      set(LIBPAM_HAVE_CONFDIR TRUE)
+    endif()
+  else()
+    set(LIBPAM_FOUND FALSE)
+  endif()
+
+  set(LIBPAM_INCLUDE_DIRS ${LIBPAM_INCLUDE_DIR})
+  set(LIBPAM_LIBRARIES ${LIBPAM_LIBRARY})
+
+  include(FindPackageHandleStandardArgs)
+  find_package_handle_standard_args(LibPAM DEFAULT_MSG LIBPAM_LIBRARIES LIBPAM_INCLUDE_DIRS)
+
+  # show the LIBPAM_INCLUDE_DIRS and LIBPAM_LIBRARIES variables only in the advanced view
+  mark_as_advanced(LIBPAM_INCLUDE_DIRS LIBPAM_LIBRARIES)
+
+endif()
--- a/src/session.c
+++ b/src/session.c
@@ -100,16 +100,16 @@ nc_gettimespec_real_add(struct timespec *ts, uint32_t msec)
 int
 nc_gettimespec_mono_add(struct timespec *ts, uint32_t msec)
 {
-    #ifdef CLOCK_MONOTONIC_RAW
+#ifdef CLOCK_MONOTONIC_RAW
     if (clock_gettime(CLOCK_MONOTONIC_RAW, ts)) {
         return -1;
     }
-    #elif defined (CLOCK_MONOTONIC)
+#elif defined (CLOCK_MONOTONIC)
     if (clock_gettime(CLOCK_MONOTONIC, ts)) {
         return -1;
     }
-    #else
-    /* no monotonic clock available, return realtime */
+#else
+    /* no monotonic clock available, return real time */
     if (nc_gettimespec_real_add(ts, 0)) {
         return -1;
     }
--- a/src/session.c
+++ b/src/session.c
@@ -57,51 +57,29 @@
 extern struct nc_server_opts server_opts;
 
 int
-nc_gettimespec_mono(struct timespec *ts)
-{
-#ifdef CLOCK_MONOTONIC_RAW
-    return clock_gettime(CLOCK_MONOTONIC_RAW, ts);
-#elif defined (CLOCK_MONOTONIC)
-    return clock_gettime(CLOCK_MONOTONIC, ts);
-#else
-    /* no monotonic clock available, return realtime */
-    return nc_gettimespec_real(ts);
-#endif
-}
-
-int
-nc_gettimespec_real(struct timespec *ts)
+nc_gettimespec_real_add(struct timespec *ts, uint32_t msec)
 {
 #ifdef CLOCK_REALTIME
-    return clock_gettime(CLOCK_REALTIME, ts);
+    if (clock_gettime(CLOCK_REALTIME, ts)) {
+        return -1;
+    }
 #else
     int rc;
     struct timeval tv;
 
     rc = gettimeofday(&tv, NULL);
-    if (!rc) {
+    if (rc) {
+        return -1;
+    } else {
         ts->tv_sec = (time_t)tv.tv_sec;
         ts->tv_nsec = 1000L * (long)tv.tv_usec;
     }
-    return rc;
 #endif
-}
 
-/* ts1 < ts2 -> +, ts1 > ts2 -> -, returns milliseconds */
-int32_t
-nc_difftimespec(const struct timespec *ts1, const struct timespec *ts2)
-{
-    int64_t nsec_diff = 0;
+    if (!msec) {
+        return 0;
+    }
 
-    nsec_diff += (((int64_t)ts2->tv_sec) - ((int64_t)ts1->tv_sec)) * 1000000000L;
-    nsec_diff += ((int64_t)ts2->tv_nsec) - ((int64_t)ts1->tv_nsec);
-
-    return nsec_diff ? nsec_diff / 1000000L : 0;
-}
-
-void
-nc_addtimespec(struct timespec *ts, uint32_t msec)
-{
     assert((ts->tv_nsec >= 0) && (ts->tv_nsec < 1000000000L));
 
     ts->tv_sec += msec / 1000;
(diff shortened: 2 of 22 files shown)


tests/pam/pam_netconf.c
```diff
@@ -0,0 +1,303 @@
...
+    free(resp);
+}
+
+/**
+ * @brief Test module's implementation of "auth" service.
+ * Prepare prompts for the client and decide based on his
+ * answers whether to allow or disallow access.
+ *
+ * @param[in] pam_h PAM handle.
+ * @param[in] flags Logically OR'd flags.
```
lePici, line 105: The OR'd flags is some term? I do not know what it means.

tests/pam/pam_netconf.c
```diff
@@ -0,0 +1,303 @@
...
+    r = nc_pam_mod_call_clb(pam_h, N_MESSAGES, msg, &resp);
+    if (r == PAM_SUCCESS) {
+        fprintf(stderr, "[TEST #5] Failed.\n");
+        r = PAM_AUTH_ERR;
+        goto cleanup;
+    }
+    printf("[TEST #5] Passed.\n\n");
+
+    /* store the correct messages */
+    msg[0] = &echo_msg;
```
lePici, line 193: Isn't that TEST 6?

tests/pam/pam_netconf.c
```diff
@@ -0,0 +1,303 @@
...
+    if (!strcmp((const char *)username, "test")) {
+        return PAM_NEW_AUTHTOK_REQD;
+    }
+    return PAM_SYSTEM_ERR;
+}
+
+/**
+ * @brief Test module's silly implementation of "password" service.
+ *
+ * @param[in] pam_h PAM handle.
```
lePici, line 268: pamh or pam_h

tests/pam/pam_netconf.c
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+cleanup:
+    return r;
+}
+
+/**
+ * @brief Test module's silly implementation of "account" service.
+ *
+ * @param[in] pam_h PAM handle.
```
lePici, line 237: pamh or pam_h?

tests/test_pam.c
```diff
@@ -0,0 +1,183 @@
...
+    nc_assert(ret & NC_PSPOLL_RPC);
+    nc_ps_clear(ps, 0, NULL);
+
+    nc_ps_free(ps);
+    nc_thread_destroy();
+    return NULL;
+}
+
+static int
+clb_hostkeys(const char *name, void *user_data, char **privkey_path, char **privkey_data,
```
lePici, line 61: Documentation would be nice.
Also in other functions.

lePici (commented):
I'm not entirely sure if the changes regarding the time functions are correct. I'll probably look at it again later.

src/session.c
```diff
@@ -56,52 +56,79 @@
...
+
+    nsec_diff += (((int64_t)ts2->tv_sec) - ((int64_t)ts1->tv_sec)) * 1000000000L;
+    nsec_diff += ((int64_t)ts2->tv_nsec) - ((int64_t)ts1->tv_nsec);
+
+    return nsec_diff ? nsec_diff / 1000000L : 0;
 }
 
 int
-nc_gettimespec_real(struct timespec *ts)
+nc_gettimespec_real_add(struct timespec *ts, uint32_t msec)
```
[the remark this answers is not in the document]
Roytak: It is extremely unlikely that a function which retrieves time fails. Therefore the return values are not checked.

src/session.c
```diff
@@ -56,52 +56,79 @@
...
-#else
-    /* no monotonic clock available, return realtime */
-    return nc_gettimespec_real(ts);
-#endif
+    int64_t nsec_diff = 0;
+
+    nsec_diff += (((int64_t)ts2->tv_sec) - ((int64_t)ts1->tv_sec)) * 1000000000L;
+    nsec_diff += ((int64_t)ts2->tv_nsec) - ((int64_t)ts1->tv_nsec);
+
+    return nsec_diff ? nsec_diff / 1000000L : 0;
```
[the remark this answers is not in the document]
jktjkt: A trivial example of them being equal is calling this function with the exact same arguments :). But the point is that the check appears to be redundant; there is no harm dividing a zero number by a large constant. It will return the exact same zero as the `: 0` part of the expression.

jktjkt (changes requested):
I only looked at the time changes so far, and I don't know how to review them because the commit message doesn't explain the changes at all.

- please explain why these changes were made; what was the problem with the old code, and how the new code is better
- please add documentation to all the new (or changed) functions

Once that's done, it will be much easier to really review the actual changes.

CMakeModules/FindLibPAM.cmake
```diff
@@ -0,0 +1,92 @@
...
+  if(LIBPAM_INCLUDE_DIR AND LIBPAM_LIBRARY)
+    set(LIBPAM_FOUND TRUE)
+
+    # there is no reliable way to check the version of PAM, so see if the declaration of a function pam_start_confdir is
+    # in the header file (added in PAM 1.4) and only then enable PAM tests
+    file(STRINGS ${LIBPAM_INCLUDE_DIR}/security/pam_appl.h PAM_CONFDIR REGEX "pam_start_confdir")
+
+    if ("${PAM_CONFDIR}" STREQUAL "")
+      # it's not there, disable PAM tests
+      message(STATUS "LibPAM found, version < 1.4, disabled PAM tests")
```
jktjkt, line 72: This doesn't belong to a `FindXYZ.cmake` module. In CMake, these files should "just" look for a library, determine its version, and set the appropriate variables. Then, in the CMake code which *requests* a module, the developer implements all the "business logic" such as only enabling the test suite when PAM is new enough.

src/session.c
```diff
@@ -116,6 +143,21 @@ nc_addtimespec(struct timespec *ts, uint32_t msec)
...
+{
+    struct timespec cur;
+    int64_t nsec_diff = 0;
+
+    nc_gettimespec_mono_add(&cur, 0);
+
+    nsec_diff += (((int64_t)ts->tv_sec) - ((int64_t)cur.tv_sec)) * 1000000000L;
+    nsec_diff += ((int64_t)ts->tv_nsec) - ((int64_t)cur.tv_nsec);
+
+    return nsec_diff ? nsec_diff / 1000000L : 0;
```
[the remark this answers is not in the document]
Roytak: It is possible, simplified the return statement to `return nsec_diff / 1000000L`, since if nsec_diff is 0, the result will still be 0.

src/session_server_ssh.c
```diff
@@ -907,32 +929,243 @@ nc_sshcb_auth_password(struct nc_session *session, ssh_message msg)
...
+            return PAM_CONV_ERR;
+        }
+    }
+
+    /* there are no requests left for the user, only messages with some information for the client were sent */
+    if (n_requests <= 0) {
+        return PAM_SUCCESS;
+    }
+
+    *resp = calloc(n_requests, sizeof **resp);
```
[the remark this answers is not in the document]
Roytak: The memory is freed by the caller as per Linux PAM's documentation : 

> It is the caller's responsibility to release both, this array and the responses
       themselves, using free(3).

Roytak:
> * please explain why these changes were made; what was the problem with the old code, and how the new code is better

While writing a section of code, which blocks and loops until a timeout is reached in order to parse a message without having to rely on a callback (authentication reasons), I noticed that these two parts can be simplified: 
` if (opts->auth_timeout) {
        nc_gettimespec_mono(&ts_timeout);
        nc_addtimespec(&ts_timeout, opts->auth_timeout * 1000);
   }` and 
 `if (opts->auth_timeout) {
            nc_gettimespec_mono(&ts_cur);
            if (nc_difftimespec(&ts_cur, &ts_timeout) < 1) {...`
  and that this pattern repeats on a lot of places. After the changes, the former can be written as:
 `if (opts->auth_timeout) {
        nc_gettimespec_mono_add(&ts_timeout, opts->auth_timeout * 1000);
    }` and
    `if ((opts->auth_timeout) && (nc_difftimespec_cur(&ts_timeout) < 1)) {...`.
    This meant the removal of functions, which simply get the current time, by functions which get the current time and may add some time to it.
     The function `int32_t nc_difftimespec(const struct timespec *ts1, const struct timespec *ts2);` remains, since it still might find some use in a place, where there is a need to get the difference of two timespec structures and neither of them holding the current time, however it is way more common to check by how much the specific time is different from the current time so a new function was added, which only accepts one argument.

> * please add documentation to all the new (or changed) functions

Done.

tests/test_pam.c
```diff
@@ -0,0 +1,183 @@
...
+#include <pthread.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <libyang/libyang.h>
+#include <log.h>
+#include <session_client.h>
+#include <session_server.h>
+#include "tests/config.h"
```
michalvasko, line 25: We always put a newline before local headers.

tests/test_pam.c
```diff
@@ -0,0 +1,183 @@
...
+    printf("SSH client finished.\n");
+    nc_client_destroy();
+
+    nc_session_free(session, NULL);
+    nc_thread_destroy();
+    return NULL;
+}
+
+int
+main()
```
michalvasko, line 140: Never use empty argument list, put `void` instead.

michalvasko (changes requested):

src/session_server_ssh.c
```diff
@@ -907,33 +936,244 @@ nc_sshcb_auth_password(struct nc_session *session, ssh_message msg)
...
+            echo[j++] = 1;
+        }
+        if (msg[i]->msg_style == PAM_PROMPT_ECHO_OFF) {
+            /* no need to set to 0 because of calloc */
+            j++;
+        }
+    }
+
+    /* print all the keyboard-interactive challenges to the user */
+    ssh_message_auth_interactive_request(clb_data->msg, name, instruction, n_requests, prompts, echo);
```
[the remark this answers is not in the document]
Roytak: Done.

src/session_server_ssh.c
```diff
@@ -907,32 +929,243 @@ nc_sshcb_auth_password(struct nc_session *session, ssh_message msg)
...
+
+    if (opts->auth_timeout) {
+        nc_gettimespec_mono_add(&ts_timeout, opts->auth_timeout * 1000);
+    }
+
+    /* get user's replies */
+    while (1) {
```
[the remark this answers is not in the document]
Roytak: Done.

src/session_server_ssh.c
```diff
@@ -907,33 +936,244 @@ nc_sshcb_auth_password(struct nc_session *session, ssh_message msg)
...
+        }
+    }
+
+cleanup:
+    /* destroy the PAM context */
+    if (pam_end(pam_h, ret) != PAM_SUCCESS) {
+        ERR(NULL, "%s", pam_strerror(pam_h, ret));
+    }
+    free(server_opts.conf_name);
+    free(server_opts.conf_dir);
```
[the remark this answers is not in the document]
Roytak: Moved to `nc_server_destroy`.

tests/pam/pam_netconf.c
Roytak: Flags can be passed to the functions an change their behavior. One of these flags can be logically OR'd with PAM_SILENT, which is used to inform the module to not pass any text (errors or warnings) to the application. I rewrote it to just flags.

src/session_p.h
```diff
@@ -520,13 +520,13 @@ struct passwd *nc_getpwuid(uid_t uid, struct passwd *pwd_buf, char **buf, size_t
 
 NC_MSG_TYPE nc_send_msg_io(struct nc_session *session, int io_timeout, struct lyd_node *op);
 
-int nc_gettimespec_mono(struct timespec *ts);
+int32_t nc_difftimespec(const struct timespec *ts1, const struct timespec *ts2);
```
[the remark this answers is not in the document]
Roytak: Done.

src/session_p.h
```diff
@@ -514,6 +516,15 @@ struct nc_ntf_thread_arg {
     void (*notif_clb)(struct nc_session *session, const struct lyd_node *envp, const struct lyd_node *op);
 };
 
+#ifdef NC_ENABLED_SSH
+
+/* arguments for a PAM callback */
```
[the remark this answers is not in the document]
Roytak: Done.

src/session_server_ssh.c
```diff
@@ -907,32 +929,243 @@ nc_sshcb_auth_password(struct nc_session *session, ssh_message msg)
...
+    ret = pam_authenticate(pam_h, 0);
+    if (ret == PAM_SUCCESS) {
+        /* correct token entered, check other requirements(the time of the day, expired token, ...) */
+        ret = pam_acct_mgmt(pam_h, 0);
+    } else {
+        ERR(NULL, "%s", pam_strerror(pam_h, ret));
+        goto cleanup;
+    }
+
+    if (ret != PAM_SUCCESS) {
```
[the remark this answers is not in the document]
Roytak: Done.

tests/pam/pam_netconf.c
```diff
@@ -0,0 +1,303 @@
...
+ * @param[in] eq_ans User's response to the second challenge.
+ * @return PAM_SUCCESS on success;
+ * @return PAM_AUTH_ERR whenever the user's replies are incorrect.
+ */
+static int
+nc_pam_mod_auth(const char *username, char *reversed_username, char *eq_ans)
+{
+    int i, j = 0;
+    int len = strlen(reversed_username);
+    char *buffer = calloc(len + 1, sizeof *buffer);
```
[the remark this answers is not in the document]
Roytak: Done.

src/session_server_ssh.c
```diff
@@ -907,33 +936,244 @@ nc_sshcb_auth_password(struct nc_session *session, ssh_message msg)
...
+    n_answers = ssh_userauth_kbdint_getnanswers(libssh_session);
+    if (n_answers != n_requests) {
+        WRN(NULL, "Expected %d response(s), got %d.", n_requests, n_answers);
+        r = PAM_CONV_ERR;
+        goto cleanup;
+    }
+
+    /* give the replies to a PAM module */
+    for (i = 0; i < n_answers; i++) {
+        (*resp)[i].resp = strdup(ssh_userauth_kbdint_getanswer(libssh_session, i));
```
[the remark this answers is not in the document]
Roytak: Done.

src/session_server_ssh.c
```diff
@@ -907,33 +948,243 @@ nc_sshcb_auth_password(struct nc_session *session, ssh_message msg)
...
+    /* initialize PAM and see if the given configuration file exists */
+#ifdef LIBPAM_HAVE_CONFDIR
+    /* PAM version >= 1.4 */
+    ret = pam_start_confdir(server_opts.conf_name, session->username, &conv, server_opts.conf_dir, &pam_h);
+#else
+    /* PAM version < 1.4 */
+    ret = pam_start(server_opts.conf_name, session->username, &conv, &pam_h);
+#endif
+    if (ret != PAM_SUCCESS) {
+        fprintf(stderr, "PAM error occurred (%s).\n", pam_strerror(pam_h, ret));
```
michalvasko, line 1141: Always use `ERR` macro, never `fprintf` directly.

tests/CMakeLists.txt
```diff
@@ -9,6 +9,12 @@ endif()
 
 # list of all the tests in each directory
 set(tests test_io test_fd_comm test_init_destroy_client test_init_destroy_server test_client_thread test_thread_messages)
+
+# only enable PAM tests if the version of PAM is greater than 1.4
+if (LIBPAM_HAVE_CONFDIR)
```
michalvasko, line 14: In `cmake`, we do not put spaces after `if` or any other similar statements.

tests/CMakeLists.txt
```diff
@@ -76,3 +82,20 @@ endif()
...
+    add_library(pam_netconf SHARED ${CMAKE_SOURCE_DIR}/tests/pam/pam_netconf.c)
+    target_link_libraries(pam_netconf ${LIBPAM_LIBRARIES})
+
+    #generate PAM configuration file
+    FILE(WRITE ${CMAKE_SOURCE_DIR}/tests/pam/netconf.conf
+    "#%PAM-1.4\n"
+    "auth required ${CMAKE_SOURCE_DIR}/build/tests/pam_netconf.so\n"
+    "account required ${CMAKE_SOURCE_DIR}/build/tests/pam_netconf.so\n"
+    "password required ${CMAKE_SOURCE_DIR}/build/tests/pam_netconf.so\n"
+    )
```
michalvasko, line 99: When writing a command on several lines, always increase the indent on the other lines. Also, `cmake` is case insensitive and we use the lowercase command names, so use just `file`.

michalvasko (changes requested):

src/session_client_ssh.c
```diff
@@ -1316,9 +1316,6 @@ connect_ssh_session(struct nc_session *session, struct nc_client_ssh_opts *opts,
                         break;
                     }
 
-                    /* libssh BUG - echo is always 1 for some reason, assume always 0 */
```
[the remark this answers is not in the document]
Roytak: This bug was fixed in libssh 0.9.0.

src/session_server_ssh.c
```diff
@@ -176,6 +177,41 @@ nc_server_ssh_set_interactive_auth_clb(int (*interactive_auth_clb)(const struct
...
+#ifdef LIBPAM_HAVE_CONFDIR
+        server_opts.conf_dir = strdup(conf_dir);
+        if (!(server_opts.conf_dir)) {
+            ERRMEM;
+            r = -1;
+        }
+#else
+        ERR(NULL, "Failed to set PAM config directory because of old version of PAM. "
+                "Put the config file in the system directory (usually /etc/pam.d/).");
+        r = -1;
```
michalvasko, line 208: In case user sets both `conf_name` and `conf_dir` and this code is executed, the return value would be -1 despite `conf_name` being set. Such behavior should be avoided, the general rule (that has almost no exceptions) is that either function fully succeeds or fails without changing anything. Partial success/fail is confusing and usually not intentional.

src/session_server_ssh.c
```diff
@@ -907,33 +948,248 @@ nc_sshcb_auth_password(struct nc_session *session, ssh_message msg)
...
+            break;
+        }
+
+        reply = ssh_message_get(libssh_session);
+        if (reply) {
+            break;
+        }
+
+        usleep(NC_TIMEOUT_STEP);
+    } while ((opts->auth_timeout) && (nc_difftimespec_cur(&ts_timeout) >= 1));
```
michalvasko, line 1070: There seem to be no error message on timeout while there is one for other conditions. My suggestion would be to treat error on line 1060 explicitly and treat condition on 1072 as timeout only.

tests/CMakeLists.txt
```diff
@@ -76,3 +82,20 @@ endif()
 
 include_directories(${CMAKE_SOURCE_DIR}/src ${PROJECT_BINARY_DIR})
 configure_file("${PROJECT_SOURCE_DIR}/tests/config.h.in" "${PROJECT_BINARY_DIR}/tests/config.h" ESCAPE_QUOTES @ONLY)
+
+if(LIBPAM_HAVE_CONFDIR)
+
+    #compile PAM test module
+    set(CMAKE_SHARED_LIBRARY_PREFIX  )
```
michalvasko, line 89: This removes the library prefix globally for the whole CMake list and any nested ones. While being fine in this case, it could cause problems later and generally it is best to set similar options per-target. Use `SET_TARGET_PROPERTIES(name PROPERTIES PREFIX "")` with `name` being the target.

michalvasko (changes requested):

src/session_server_ssh.c
Roytak: `conf_dir` is now set first. This means that if it fails, `conf_name` remains unchanged.

michalvasko (approved):

Issue #389, opened 2022-07-26
PAM test failure with linux-pam-1.5.2


jktjkt:
I'm on the latest `devel` of everything, plus #388:
```
10/13 Test  #8: test_pam .........................Subprocess aborted***Exception:   0.10 sec
[INF]: Listening on 0.0.0.0:6002 for SSH connections.
[INF]: Trying to connect via IPv4 to 0.0.0.0:6002.
[INF]: Successfully connected to 0.0.0.0:6002 over IPv4.
[INF]: Accepted a connection on 0.0.0.0:6002 from 127.0.0.1:42736.
[INF]: Received an SSH message "request-service" of subtype "ssh-userauth".
[INF]: Received an SSH message "request-auth" of subtype "none".
[INF]: Keyboard-interactive authentication.
[INF]: Received an SSH message "request-auth" of subtype "interactive".
[INF]: PAM error occurred (Module is unknown).

[INF]: Failed user "test" authentication attempt (#1).
[WRN]: Authentication denied.
[ERR]: Unable to authenticate to the remote server (all attempts via supported authentication methods failed).
assert failed (/home/jkt/work/cesnet/gerrit/github/CESNET/libnetconf2/tests/test_pam.c:132)
```

I do not have any systemwide PAM configuration for `libnetconf2`, and the test suite runs as an unprivileged user.

Cc: @Roytak

jktjkt:
OK, that was easy enough, #390 fixes this one.

PR #390, opened 2022-07-26
tests: do not make assumptions about build directory layout

jktjkt:
The tests would only work if the build directory was named `build/`, and if it was a subdirectory of the source tree.

Bug: #389
Fixes: 41a11e4 pam UPDATE auth using Linux PAM

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -93,9 +93,9 @@ if(LIBPAM_HAVE_CONFDIR)
     #generate PAM configuration file
     file(WRITE ${CMAKE_SOURCE_DIR}/tests/pam/netconf.conf
         "#%PAM-1.4\n"
-        "auth required ${CMAKE_SOURCE_DIR}/build/tests/pam_netconf.so\n"
-        "account required ${CMAKE_SOURCE_DIR}/build/tests/pam_netconf.so\n"
-        "password required ${CMAKE_SOURCE_DIR}/build/tests/pam_netconf.so\n"
+        "auth required ${CMAKE_CURRENT_BINARY_DIR}/pam_netconf.so\n"
+        "account required ${CMAKE_CURRENT_BINARY_DIR}/pam_netconf.so\n"
+        "password required ${CMAKE_CURRENT_BINARY_DIR}/pam_netconf.so\n"
     )
 
 endif()


tests/CMakeLists.txt
```diff
@@ -93,9 +93,9 @@ if(LIBPAM_HAVE_CONFDIR)
     #generate PAM configuration file
     file(WRITE ${CMAKE_SOURCE_DIR}/tests/pam/netconf.conf
         "#%PAM-1.4\n"
-        "auth required ${CMAKE_SOURCE_DIR}/build/tests/pam_netconf.so\n"
-        "account required ${CMAKE_SOURCE_DIR}/build/tests/pam_netconf.so\n"
-        "password required ${CMAKE_SOURCE_DIR}/build/tests/pam_netconf.so\n"
+	"auth required ${CMAKE_CURRENT_BINARY_DIR}/pam_netconf.so\n"
+	"account required ${CMAKE_CURRENT_BINARY_DIR}/pam_netconf.so\n"
+	"password required ${CMAKE_CURRENT_BINARY_DIR}/pam_netconf.so\n"
```
Roytak, line 98: indent as above

tests/pam/netconf.conf
```diff
@@ -0,0 +1,4 @@
+#%PAM-1.4
+auth required /home/jkt/work/prog/_build/czechlight-clang14-asan-ubsan-ly2/libnetconf2/tests/pam_netconf.so
+account required /home/jkt/work/prog/_build/czechlight-clang14-asan-ubsan-ly2/libnetconf2/tests/pam_netconf.so
+password required /home/jkt/work/prog/_build/czechlight-clang14-asan-ubsan-ly2/libnetconf2/tests/pam_netconf.so
```
Roytak, line 4: remove the file from the commit

Roytak (changes requested):

tests/pam/netconf.conf
jktjkt: Oops, I was too lazy with the `git add`. That's the danger of generating files in the src dir :).

tests/CMakeLists.txt
jktjkt: sorry, good catch

Roytak (approved):

PR #412, opened 2023-03-23
fix of crash when a connection is made in background

tieliao:
working in progress

### DIFF
--- a/src/session_client.c
+++ b/src/session_client.c
@@ -1634,6 +1634,8 @@ nc_sock_connect(const char *host, uint16_t port, int timeout_ms, struct nc_keepa
     struct addrinfo hints, *res_list = NULL, *res;
     char *buf, port_s[6]; /* length of string representation of short int */
     void *addr;
+    struct sockaddr saddr;
+    socklen_t addr_len;
 
     DBG(NULL, "nc_sock_connect(%s, %u, %d, %d)", host, port, timeout_ms, sock);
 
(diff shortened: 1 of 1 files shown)


michalvasko:
Could you maybe provide more information? `valgrind` output for the crash maybe? I will try to fix it then, you do not have to.

tieliao:
A segmentation fault occurred when I make 10 successful callhome connections to a callhome server, shutdown and restart the server. The backtrace at free(session->host):
#0  __GI___libc_free (mem=0x6) at malloc.c:3093
#1  nc_session_free (session=0x7fffd0000b90, data_free=0x0) at /home/user/share/libnetconf2/src/session.c:997
#2  nc_connect_ch_endpt (endpt=0x555555805b80, acquire_ctx_cb=0x5555555772a2 <sim_session_acquire_ctx_cb>, release_ctx_cb=0x5555555772e6 <sim_session_release_ctx_cb>, ctx_cb_data=0x5555557d9c60, session=0x7fffdf66de30) at /home/user/share/libnetconf2/src/session_server.c:3405
#3  nc_ch_client_thread (arg=0x5555557fbfc0) at /home/user/share/libnetconf2/src/session_server.c:3555
#4  start_thread (arg=<optimized out>) at pthread_create.c:486
#5  clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

michalvasko:
Oh, the `ip_host` is not being set in case a pending socket is used. Thanks for the fix but there is some code duplication (that existed even before your changes) so you may try to fix that. However, I can do that, too, after I merge your PR so at least please change the target branch to `devel`, we do not accept PRs to `master`.

tieliao:
OK I will switch to the devel branch the next time and let you fix this issue. FYI I'm writing a callhome simulator that can simulate thousands of netconf sessions. I've also fixed an issue on the limit of the number of connections to 1019. This is due to pthread_key_create(), as discussed in another thread. The fix is in libyang. Would you like that I submit my changes?
Besides I'm facing a memory leak issue after 1k connections are made and running for several hours. I did not yet have time to look at it.

michalvasko:
> OK I will switch to the devel branch the next time and let you fix this issue.

Fine, merged to `devel`.

> I've also fixed an issue on the limit of the number of connections to 1019. This is due to pthread_key_create(), as discussed in another thread. The fix is in libyang. Would you like that I submit my changes?

Please do, I am interested in seeing your solution because I could not think if any keeping all the features it provided.

> Besides I'm facing a memory leak issue after 1k connections are made and running for several hours. I did not yet have time to look at it. So it could be a problem at my side.

`valgrind` output can be helpful or just using the memory sanitizer as well.

> Another issue is that my simulator can not make more than 1725 netconf sessions, this is strange to me and I still don't understand.

Cannot help you without more details.

PR #425, opened 2023-06-19
session BUGFIX get source host correctly on a pending socket

barbarost:
`getpeername()` needs to be used to get remote host's address info instead of `getsockname()`.


michalvasko (approved):

michalvasko:
I think you are right, merged.

PR #436, opened 2023-08-31
Adding initial pkcs#11 support for TLS private key storage

JohnLogan:
This is an initial commit adding pkcs#11 support for TLS private key storage. This was tested using Intel 'Key Management Reference Architecture (KMRA)' project, which uses SGx enclave to store manage keys, but could be used with other pkcs#11 targets. Moves private keys from local fiel to remote key management storage.

### DIFF
--- a/src/session_client_tls.c
+++ b/src/session_client_tls.c
@@ -22,6 +22,7 @@
 #include <errno.h>
 #include <string.h>
 #include <unistd.h>
+#include <openssl/engine.h>
 
 #include <libyang/libyang.h>
 #include <openssl/err.h>
@@ -508,6 +509,10 @@ nc_client_tls_update_opts(struct nc_client_tls_opts *opts, const char *peername)
     char *key;
     X509_LOOKUP *lookup;
     X509_VERIFY_PARAM *vpm = NULL;
+    const int CMD_MANDATORY = 0;
+    EVP_PKEY *pkey = NULL;
+    ENGINE * pkcs11 = NULL;
+    const char* opensc_pkcs11_so = getenv("MODULE");
 
     if (!opts->tls_ctx || opts->tls_ctx_change) {
         SSL_CTX_free(opts->tls_ctx);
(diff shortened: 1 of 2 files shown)


src/session_client_tls.c
```diff
@@ -540,13 +545,57 @@ nc_client_tls_update_opts(struct nc_client_tls_opts *opts, const char *peername)
...
+        if ( pkcs11 == NULL )
+        {
+            ERR(NULL, "Error retrieving 'pkcs11' engine");
+            rc = -1;
+            goto cleanup;
+        }
+
+        if ( 0 != access( opensc_pkcs11_so, R_OK ) )
+        {
+            ERR(NULL, "Error finding '/usr/local/lib/libpkcs11-proxy.so'");
```
jktjkt, line 560: paths cannot be hardcoded like that, not even in error messages

src/session_client_tls.c
```diff
@@ -540,13 +545,57 @@ nc_client_tls_update_opts(struct nc_client_tls_opts *opts, const char *peername)
...
             rc = -1;
             goto cleanup;
         }
 
+        if ( 1 != ENGINE_ctrl_cmd_string( pkcs11, "MODULE_PATH", opensc_pkcs11_so, CMD_MANDATORY ) )
```
jktjkt, line 565: This loads a `.so` into the process' address space based on a an environment variable. That's effectively an arbitrary code execution vulnerability based on a non-standard variable name.

How does OpenSSL typically solve this?

Also, [the README upstream](https://github.com/openssl/openssl/blob/master/README-ENGINES.md) says that the command name is called `SO_PATH`, not `MODULE_PATH`.

src/session_client_tls.c
```diff
@@ -540,13 +545,57 @@ nc_client_tls_update_opts(struct nc_client_tls_opts *opts, const char *peername)
...
-            ERR(NULL, "Loading the client private key from \'%s\' failed (%s).", key,
-                    ERR_reason_error_string(ERR_get_error()));
+
+        ENGINE_load_dynamic();
+        pkcs11 = ENGINE_by_id( "pkcs11" );
+        if ( pkcs11 == NULL )
+        {
+            ERR(NULL, "Error retrieving 'pkcs11' engine");
+            rc = -1;
+            goto cleanup;
```
jktjkt, line 555: This means that the code would hard-fail at runtime when OpenSSL is built without this dynamic engine support. It's better to only request a dynamic engine when needed, i.e., hide this behind some switch to active this new functionality.

The library should continue to work as-is when the user does not opt-in to this new feature (which by definition requires some configuration).

src/session_client_tls.c
```diff
@@ -540,13 +545,57 @@ nc_client_tls_update_opts(struct nc_client_tls_opts *opts, const char *peername)
...
+        }
+
+        if ( 1 != ENGINE_init( pkcs11 ) )
+        {
+		    ERR(NULL, "Error pkcs11: unable to initialize engine");
+            rc = -1;
+            goto cleanup;
+        }
+
+        if ( 1 != ENGINE_ctrl_cmd_string( pkcs11, "PIN", "1234", CMD_MANDATORY ) )
```
jktjkt, line 579: a hardcoded PIN?

src/session_server_tls.c
```diff
@@ -1781,26 +1787,70 @@ nc_tls_ctx_set_server_cert_key(SSL_CTX *tls_ctx, const char *cert_name)
...
+        }
+
+        if ( 1 != ENGINE_ctrl_cmd_string( pkcs11, "MODULE_PATH", opensc_pkcs11_so, CMD_MANDATORY ) )
+        {
+            ERR(NULL, "Error setting module_path");
+            ret = -1;
+            goto cleanup;
+        }
+
+        if ( 1 != ENGINE_init( pkcs11 ) )
```
jktjkt, line 1821: The documentation says:

> This returns zero if the ENGINE was not already operational and couldn't be successfully initialised (e.g. lack of system drivers, no special hardware attached, etc), otherwise it will return nonzero to indicate that the ENGINE is now operational and will have allocated a new functional reference to the ENGINE.

...yet the code checks for `1`.

Also everywhere else.

jktjkt (changes requested):

vjardin:
another related merge request is the !454 

Since ENGINE is deprecated since OpenSSL 3, PROVIDER shall be used instead.

PR #446, opened 2023-11-14
Support building a static library

jktjkt:
This is literally the bare minimum that one has to do in order to let a properly configured CMake do its thing and produce a static library.

*v2: be lazy and rely on CMake's builtin features*

~~I'm deliberately not using CMake's standard way through the `BUILD_SHARED_LIBS` option because the other libraries in the NETCONF/YANG stack already stick to this custom convention.~~


michalvasko (approved):

michalvasko:
Alright, seems like the right way to support this and we should update other projects this way, too. But please fix the conflicts before merging.

PR #453, opened 2024-01-19
server_config BUGFIX remove delete unix endpoint

jeremie6wind:
The unix endpoint can be started with -U option of netopeer2-server. However it is not in the yang config. Leave it when deleting other endpoints.


Roytak:
Hi, this makes sense, however it will require more work than this. Now for example the netopeer2-server's UNIX socket will never be deleted. But thank you, I'll merge it and complete the bug fix, once I've tested it.

PR #454, opened 2024-01-19
client and server tls UPDATE add pkcs11 engine support

vkarpenk:
This PR is adding pkcs#11 support for TLS private key storage. This was tested using Intel 'Key Management Reference Application (KMRA)' project, which uses SGX enclave to store manage keys, but could be used with other pkcs#11 targets. Moves private keys from local file to remote key management storage.

### DIFF
--- a/src/session_client_tls.c
+++ b/src/session_client_tls.c
@@ -27,6 +27,7 @@
 #include <openssl/err.h>
 #include <openssl/ossl_typ.h>
 #include <openssl/x509.h>
+#include <openssl/engine.h>
 
 #include "config.h"
 #include "log_p.h"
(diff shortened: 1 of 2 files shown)


vjardin:
Since ENGINE is deprecated with OpenSSL 3, please, switch to the PROVIDER API.

src/session_server_tls.c
```diff
@@ -862,6 +863,9 @@ nc_tls_ctx_set_server_cert_key(SSL_CTX *tls_ctx, struct nc_server_tls_opts *opts
     NC_PRIVKEY_FORMAT privkey_type;
     X509 *cert = NULL;
     EVP_PKEY *pkey = NULL;
+    ENGINE *pkcs11 = NULL;
+    const char* uri = getenv("TOKEN_KEY_URI");
+    const char* pin = getenv("DEFAULT_USER_PIN");
```
jktjkt, line 868: Are these standard variable names, and are they documented somewhere?

src/session_client_tls.c
```diff
@@ -387,6 +388,9 @@ nc_client_tls_update_opts(struct nc_client_tls_opts *opts, const char *peername)
     char *key;
     X509_LOOKUP *lookup;
     X509_VERIFY_PARAM *vpm = NULL;
+    EVP_PKEY *pkey = NULL;
+    ENGINE *pkcs11 = NULL;
+    const char* pin = getenv("DEFAULT_USER_PIN");
```
jktjkt, line 393: Is this a standard variable name used by other projects, and does it conform to how the OpenSSL API is "supposed to be used"? Where is the documentation?

jktjkt (changes requested):
This really needs docs. Please link to an some best-practices document or a HOWTO by the OpenSSL project which explains how these are used in other SW.

How is the end user supposed to control this feature, btw?

PR #475, opened 2024-04-08
Parallel tests

jktjkt:
Running tests in parallel was broken because these files all had some hard-coded port numbers. Some of these port numbers were OK, others were overlapping. Let's fix that by letting CMake manage these port numbers. However, to do that, one has to first convert the CMake setup to a "more idiomatic" approach with one function call per test. As a bonus, this gets rid of some "random variables" and the final test declarations are, hopefully, a bit easier to read. See the individual commits for descriptions.


Roytak (approved):

michalvasko:
Okay, thanks, merging.

PR #487, opened 2024-05-22
Support SSH authentication for non-configured users

### LINKED ISSUE #486: Authenticate users with central user management
I have been using netopeer2 as netconf server and the users are described in ietf-system and becomes entries in /etc/shadow,
This has worked before, but now (since the ietf-netconf-server YANG upgrade and libnetconf refactor),
it is also required to specify them in in ietf-netconf-server  as well. 

<pre>
{
    "ietf-netconf-server:netconf-server": {
	"listen": {
	    "endpoints": {
		"endpoint": [
		    {
			"name": "default-ssh",
			"ssh": {
			    "tcp-server-parameters": {
				"local-address": "::"
			    },
			    "ssh-server-parameters": {
				"server-identity": {
				    "host-key": [
					{
					    "name": "default-key",
					    "public-key": {
						"central-keystore-reference": "genkey"
					    }
					}
				    ]
				},
<b>
				"client-authentication": {
				    "users": {
					"user": [
					    {
						"name": "admin",
						"libnetconf2-netconf-server:keyboard-interactive": {
						    "libnetconf2-netconf-server:use-system-auth": {}
						}
					    }
					]
				    }
</b>
				}
			    }
			}
		    }
		]
	    }
	}
    }
}
</pre>
Why this change? Or am I doing something wrong?

Roytak:
Based on the value of the YANG feature `local-users-supported` the SSH server decides, whether to lookup authorized users from the configuration or from the system.

Resolves #486

### DIFF
--- a/src/session_server_ssh.c
+++ b/src/session_server_ssh.c
@@ -48,8 +48,15 @@
 
 extern struct nc_server_opts server_opts;
 
+/**
+ * @brief Stores the private key data as a temporary file.
+ *
+ * @param[in] in Private key data.
+ * @param[in] privkey_format String representation of the private key format.
+ * @return Path to the created temporary file or NULL on fail.
+ */
 static char *
-base64der_privkey_to_tmp_file(const char *in, const char *privkey_format)
+nc_server_ssh_privkey_data_to_tmp_file(const char *in, const char *privkey_format)
 {
     char path[12] = "/tmp/XXXXXX";
     int fd, written;
@@ -110,6 +117,13 @@ base64der_privkey_to_tmp_file(const char *in, const char *privkey_format)
     return strdup(path);
 }
 
+/**
+ * @brief Get asymmetric key from the keystore.
+ *
+ * @param[in] referenced_name Name of the asymmetric key in the keystore.
+ * @param[out] askey Referenced asymmetric key.
+ * @return 0 on success, 1 on error.
+ */
 static int
 nc_server_ssh_ks_ref_get_key(const char *referenced_name, struct nc_asymmetric_key **askey)
 {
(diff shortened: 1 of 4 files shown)


src/session_server_ssh.c
```diff
@@ -468,20 +468,167 @@ nc_server_ssh_get_system_keys(const char *username, struct nc_public_key **pubke
...
+    struct spwd *spwd, spwd_buf;
+    char *pass_hash = NULL, *buf = NULL;
+    size_t buf_size = 256;
+
+    buf = malloc(buf_size);
+    NC_CHECK_ERRMEM_GOTO(!buf, , error);
+
+    pwd = nc_server_ssh_getpwnam(username, &pwd_buf, &buf, &buf_size);
+    if (!pwd) {
+        VRB(NULL, "User \"%s\" not found locally.", username);
```
jktjkt, line 572: The YANG-level feature which enables this behavior is called `local-users-supported`, and if it's enabled, then "local" refers to YANG config. Please change this string so that it says "in the system configuration" or "outside of YANG data" or something, but let's not call them "local".

Also everywhere else, e.g., the docstring for `nc_server_ssh_auth_kbdint_system` and the local variables in there.

jktjkt (changes requested):
Thanks for this, I'm thrilled to see that the "system" behavior is back now.

src/session_server_ssh.c
Roytak: Right, changed these occurrences to contain the word `system`.

src/session_server_ssh.c
```diff
@@ -403,20 +468,167 @@ nc_server_ssh_get_system_keys(const char *username, struct nc_public_key **pubke
...
+        r = getspnam_r(username, spwd_buf, *buf, *buf_size, &spwd);
+# else
+        spwd = getspnam_r(username, spwd_buf, *buf, *buf_size);
+# endif
+        if (spwd) {
+            /* entry found */
+            break;
+        }
+
+        if (r == ERANGE) {
```
michalvasko, line 538: On different platforms than QNX `r` will not be set and always be `0`.

michalvasko (changes requested):
Except for my comment, I would add a fairly brief section into README called `Supported YANG modules` or something similar mentioning what YANG modules with what features the library expects to be implemented and that there is a function that creates such a context. Mention also that this single feature can be disabled and also an explicit list of the YANG modules and their features, maybe aslo with a list of disabled features of all the modules. That way a concise summary of supported features of *libnetconf2* will be included and can be updated as new features are supported.

michalvasko (approved):

PR #490, opened 2024-07-03
Remove CRL path and url configuration

Roytak:
CRLs are now obtainable only through their x509 extension. This PR removes the following APIs:
- nc_client_tls_set_crl_paths
- nc_client_tls_get_crl_paths
- nc_client_tls_ch_set_crl_paths
- nc_client_tls_ch_get_crl_paths


michalvasko (changes requested):
Leave the functions there but mark them deprecated and leave them empty (not doing anything). That will keep the ABI compatibility.

michalvasko (approved):

PR #493, opened 2024-07-10
Add support for certificate expiration notifications

Roytak:
This PR implements the logic of the `certificate-expiration-notification` YANG feature, which is still not loaded by default, but if it is now enabled, then the server will start generating such notifications. The times of these notifications can be customized with YANG data.


src/session_client.c
```diff
@@ -2452,10 +2452,12 @@ nc_recv_notif_thread(void *arg)
...
             notif_clb(session, envp, op, user_data);
             if (!strcmp(op->schema->name, "notificationComplete") && !strcmp(op->schema->module->name, "nc-notifications")) {
-                lyd_free_tree(envp);
+                lyd_free_all(envp);
+                lyd_free_all(op);
                 break;
             }
-            lyd_free_tree(envp);
+            lyd_free_all(envp);
+            lyd_free_all(op);
```
Roytak, line 2460: Found a memory leak here when using `netopeer2-cli` to test this. Not sure if this is the right way to free the memory here.

src/session_p.h
```diff
@@ -514,6 +579,22 @@ struct nc_server_opts {
...
+
+    /**
+     * @brief Intervals for certificate expiration notifications.
+     */
+    struct nc_interval {
+        struct nc_cert_exp_time anchor;     /**< Lower bound of the given interval. */
+        struct nc_cert_exp_time period;     /**< Period of the given interval. */
+    } *intervals;
+    int interval_count;                     /**< Number of intervals. */
+#endif
```
Roytak, line 597: Maybe all of this should be encapsulated into a single `struct`, since all of these members are used in the same context and the names could then be shorter.

src/session_server.c
```diff
@@ -3080,3 +3098,754 @@ nc_session_get_notif_status(const struct nc_session *session)
...
+            if (r) {
+                ERR(NULL, "Joining the certificate expiration notification thread failed (%s).", strerror(r));
+            }
+        }
+    } else {
+        /* thread is not running */
+        /* UNLOCK */
+        pthread_mutex_unlock(&server_opts.cert_exp_notif_thread_lock);
+    }
+}
```
Roytak, line 3849: Not sure about this design. I wanted it to be blocking in the tests I ran, but someone may possibly not want to wait for the thread to terminate? Should the `wait` parameter be kept?

src/server_config.c
```diff
@@ -3752,6 +3752,182 @@ nc_server_config_parse_netconf_server(const struct lyd_node *node, NC_OPERATION
...
+            (server_opts.cert_exp_notif.interval_count + 1) * sizeof *server_opts.cert_exp_notif.intervals);
+    NC_CHECK_ERRMEM_RET(!server_opts.cert_exp_notif.intervals, 1);
+
+    /* convert and set the anchor */
+    ret = nc_server_config_yang_value2cert_exp_time(anchor, &cert_exp_time);
+    if (ret) {
+        goto cleanup;
+    }
+    memcpy(&server_opts.cert_exp_notif.intervals[server_opts.cert_exp_notif.interval_count].anchor,
+            &cert_exp_time, sizeof cert_exp_time);
```
michalvasko, line 3825: You can assign structs, more readable.

src/server_config.c
```diff
@@ -3752,6 +3752,182 @@ nc_server_config_parse_netconf_server(const struct lyd_node *node, NC_OPERATION
...
+        ERR(NULL, "Interval \"%s %s\" not found.", anchor, period);
+        return;
+    }
+
+    server_opts.cert_exp_notif.interval_count--;
+    if (!server_opts.cert_exp_notif.interval_count) {
+        free(server_opts.cert_exp_notif.intervals);
+        server_opts.cert_exp_notif.intervals = NULL;
+    } else if (i != server_opts.cert_exp_notif.interval_count) {
+        memcpy(&server_opts.cert_exp_notif.intervals[i],
```
michalvasko, line 3870: Assignment.

src/session_client.c
michalvasko: Seems fine except I would expect (which is probably the problem here) these 2 trees connected.

src/session_server.c
```diff
@@ -3080,3 +3098,758 @@ nc_session_get_notif_status(const struct nc_session *session)
...
+    if (server_opts.cert_exp_notif.thread_running) {
+        /* set the tid and running flag to 0, signal the thread and unlock its mutex */
+        server_opts.cert_exp_notif.thread_running = 0;
+        server_opts.cert_exp_notif.tid = 0;
+        pthread_cond_signal(&server_opts.cert_exp_notif.cond);
+
+        /* UNLOCK */
+        pthread_mutex_unlock(&server_opts.cert_exp_notif.lock);
+        if (wait) {
+            r = pthread_join(tid, NULL);
```
michalvasko, line 3843: To avoid memory leaks, if `!wait`, you need to detach the thread. I am also not sure if not just always wait until the thread finishes, but I have no strong opinion.

src/session_server.h
```diff
@@ -55,6 +55,21 @@ extern "C" {
...
+ * It is up to the user to decide what to do with the notification.
+ *
+ * In case an error occurs and you wish to terminate the notification thread,
+ * call nc_server_notif_cert_expiration_thread_stop().
+ *
+ * @param[in] expiration_time Expiration time of the certificate obtained via ly_time_time2str().
+ * @param[in] xpath Xpath of the certificate. Can be used to create the notification data.
+ * @param[in] user_data Arbitrary user data.
+ */
+typedef void(*nc_cert_exp_notif_clb)(const char *expiration_time, const char *xpath, void *user_data);
```
michalvasko, line 71: Space after `void`.

michalvasko (changes requested):

src/session_server.c
Roytak: Added the missing detach, but kept the wait parameter.

michalvasko (approved):

PR #503, opened 2024-08-14
Obtain test ports from env and share code between tests

Roytak:
Previously, the ports were set via a macro, which made a normal test and a valgrind one share the same port. This sometimes led to one of them failing when run concurrently (as in this [job](https://github.com/CESNET/libnetconf2/actions/runs/10352715968/job/28653942800) for example).

Now they are obtained from the environment that is generated by cmake.
This PR also includes the first step of sharing code between tests.


tests/CMakeLists.txt
```diff
@@ -12,70 +12,77 @@ foreach(src IN LISTS libsrc)
...
+        set(${PORT_COUNT} 1)
+    endif()
+
+    SET(PORT_INDEX 0)
+    while(PORT_INDEX LESS ${${PORT_COUNT}})
+        list(APPEND ${PORT_DEFINITIONS} "TEST_PORT_${PORT_INDEX}=${NEXT_TEST_PORT}")
+        math(EXPR PORT_INDEX "${PORT_INDEX} + 1")
+        math(EXPR NEXT_TEST_PORT "${NEXT_TEST_PORT} + 1")
+    endwhile()
+    set(NEXT_TEST_PORT ${NEXT_TEST_PORT} PARENT_SCOPE)
```
michalvasko, line 29: `PARENT_SCOPE` is probably not needed, unless I am mistaken, macros share the scope of the caller, only functions do not.

michalvasko (changes requested):

tests/CMakeLists.txt
Roytak: It is needed, because the macro is invoked inside a function and `NEXT_TEST_PORT` is a global variable. Can also define it as the function's variable and then it would not be needed, but this made more sense to me.

michalvasko: Okay, I suppose, but seeing that in a macro is always confusing.

michalvasko (approved):

PR #517, opened 2024-11-04
Add Keylog file with env variable SSLKEYLOGFILE (for openssl)

gcolin:
Hello, this PR adds the generation of keylog file for debugging TLS with wireshark. This is working fine with openssl.

The env variable SSLKEYLOGFILE is used in others apps such as Curl, Firefox and Google Chrome.

### DIFF
--- a/src/session_openssl.c
+++ b/src/session_openssl.c
@@ -21,6 +21,7 @@
 
 #include <ctype.h>
 #include <poll.h>
+#include <pthread.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


src/keylog.c
```diff
@@ -0,0 +1,72 @@
...
+#define KEYLOG_LABEL_MAXLEN (sizeof("CLIENT_HANDSHAKE_TRAFFIC_SECRET") - 1)
+
+#define CLIENT_RANDOM_SIZE  32
+
+/*
+ * The master secret in TLS 1.2 and before is always 48 bytes. In TLS 1.3, the
+ * secret size depends on the cipher suite's hash function which is 32 bytes
+ * for SHA-256 and 48 bytes for SHA-384.
+ */
+#define SECRET_MAXLEN       48
```
michalvasko, line 35: What are these defines for? They do not seem to be used anywhere.

src/session_openssl.c
```diff
@@ -43,6 +44,36 @@
 #include <openssl/x509.h>
 #include <openssl/x509v3.h>
 
+void
+tls_keylog_write_line(const SSL *ssl, const char *line)
+{
+    (void)ssl;
```
michalvasko, line 50: We use `UNUSED()` for the parameter instead.

michalvasko (changes requested):
Thanks for the PR, but to follow our code organization, a few changes will be required. 

Starting with some minor issues, we generally use the `nc_` prefix for all *libnetconf2* functions (possible exception only for static ones) to set a "namespace" differentiating between all the used libraries. But, since this is OpenSSL-only, I think it can all be put into the `session_openssl.c` file without introducing additional ones. Ideally, there should be a TLS-specific `init` function (for example), which will not do anything for the mbedTLS but initializes this keylogger for OpenSSL to follow our TLS library wrappers.

src/session_openssl.c
gcolin: ok done

gcolin:
Hello Michal, I updated the PR. the code is only in session_openssl.c

I added 2 global variable in file session_openssl.c: active_sessions and keylog_file_fp. maybe there is a better place to store these global variables.

src/session_openssl.c
```diff
@@ -43,6 +43,73 @@
 #include <openssl/x509.h>
 #include <openssl/x509v3.h>
 
+#define SSLKEYLOGFILE_ENV "SSLKEYLOGFILE"
+/* The fp for the open SSLKEYLOGFILE, or NULL if not open */
+FILE *keylog_file_fp = NULL;
+uint32_t active_sessions = 0;
+
+FILE*
```
Roytak, line 51: Use static and add the missing space, so it's `static FILE *`.

src/session_openssl.c
```diff
@@ -43,6 +43,73 @@
...
+nc_tls_keylog_open(void)
+{
+    char *keylog_file_name = getenv(SSLKEYLOGFILE_ENV);
```
Roytak, line 54: Separate the variable definition and the assignment.

src/session_openssl.c
```diff
@@ -43,6 +43,73 @@
...
+                    fclose(keylog_file_fp);
+                    keylog_file_fp = NULL;
+                }
+            }
+        }
+    }
+    return keylog_file_fp;
+}
+
+void
```
Roytak, line 71: Use static.

src/session_openssl.c
```diff
@@ -43,6 +43,73 @@
...
+nc_tls_keylog_close(void)
+{
+    char *keylog_file_name = getenv(SSLKEYLOGFILE_ENV);
```
Roytak, line 74: Separate the variable definition and the assignment.

src/session_openssl.c
```diff
@@ -43,6 +43,73 @@
...
+    if (keylog_file_name) {
+        active_sessions--;
+        if (active_sessions == 0) {
+            fclose(keylog_file_fp);
+            keylog_file_fp = NULL;
+        }
+    }
+}
+
+void
```
Roytak, line 84: Use static.

src/session_openssl.c
```diff
@@ -43,6 +43,73 @@
...
+        }
+    }
+    return keylog_file_fp;
+}
+
+void
+nc_tls_keylog_close(void)
+{
+    char *keylog_file_name = getenv(SSLKEYLOGFILE_ENV);
+    if (keylog_file_name) {
```
Roytak, line 75: Could also check the value of the `keylog_file_fp` variable, so that this variable may be redundant if I'm not mistaken.

Roytak (commented):
Hi, thank you for the contribution. Can you change the target branch to `devel`? Also would be great if you could run `make format` once you fix the small details I commented on. I think it will be good to merge then, if all tests pass.

gcolin:
There is a mutex to fix a concurrency issue in test_tls: the file was open twice.

I cannot change the "from" nokia:master. Should I recreate a new PR?

Roytak:
Is there a specific reason as to why the filename is passed as an env variable or that's just how curl handles it? I am thinking that setting it via an API may be better (and possibly doing this in netopeer2). Anyways, it will most likely be merged soon. I'll try to find out if this can be done with mbedtls as well, and possibly edit it accordingly.

gcolin:
> Is there a specific reason as to why the filename is passed as an env variable or that's just how curl handles it? I am thinking that setting it via an API may be better (and possibly doing this in netopeer2). Anyways, it will most likely be merged soon. I'll try to find out if this can be done with mbedtls as well, and possibly edit it accordingly.

Hi, yes it is simpler:
. there is only one binary to update
. curl uses it. In our software, we use both curl and netopeer2
. other popular tools use this way

A colleague also suggested to have a paramater from netopeer2. For that, netopeer2 needs also an update.

That is cool to have the same in mbedtls.

Roytak:
Hello, I decided to implement it since it required a bit more changes and I liked this enhancement. Sorry for not merging yours. Nonetheless, feel free to try it out and give us some feedback if need be. Thank you.

PR #539, opened 2025-06-11
modules UPDATE update tls modules

jeremie6wind:
iana-tls-cipher-suite-algs@2025-04-04.yang is generated thanks to the script in the Appendix using the command 'python gen-yang-modules.py'.

Link: https://datatracker.ietf.org/doc/rfc9645


jeremie6wind:
Hi,
thank you for your reply.
Let's wait then. There is nothing I need specifically.
Best regards.

PR #548, opened 2025-08-07
Coverity_fixes_jsc-se

jsc-se:
Fixes for Coverity issues (company internal).


src/io.c
```diff
@@ -719,34 +719,41 @@ nc_write_clb_flush(struct nc_wclb_arg *warg)
...
         c = nc_write_clb_flush(warg);
-        if (c == -1) {
+        /* check if c in a valid range */
+        assert((int) (ssize_t) c == c);
+        if (c < 0) {
             return -1;
         }
-        ret += c;
+        ret += (ssize_t) c;
     }
```
michalvasko, line 757: In the `proxy` branch, which will soon be merged into `master`, return values of `nc_write_clb_flush()` and `nc_write_endtag()` were changed to `int` so these changes are no longer needed.

src/server_config.c
```diff
@@ -2629,7 +2629,7 @@ nc_server_config_endpoint_reference(const struct lyd_node *node, enum nc_operati
             /* listen */
             free(endpt->referenced_endpt_name);
             endpt->referenced_endpt_name = NULL;
-        } else {
+        } else if (ch_endpt) {
```
michalvasko, line 2632: Should be an `assert()`.

src/server_config.c
```diff
@@ -2657,7 +2657,7 @@ nc_server_config_endpoint_reference(const struct lyd_node *node, enum nc_operati
             free(endpt->referenced_endpt_name);
             endpt->referenced_endpt_name = strdup(lyd_get_value(node));
             NC_CHECK_ERRMEM_GOTO(!endpt->referenced_endpt_name, ret = 1, cleanup);
-        } else {
+        } else if (ch_endpt) {
```
michalvasko, line 2660: Should be an `assert()`.

src/session_server.c
```diff
@@ -2728,6 +2728,7 @@ nc_connect_ch_endpt(struct nc_ch_endpt *endpt, nc_server_ch_session_acquire_ctx_
     sock = nc_sock_connect(endpt->src_addr, endpt->src_port, endpt->dst_addr, endpt->dst_port,
             NC_CH_CONNECT_TIMEOUT, &endpt->ka, &endpt->sock_pending, &ip_host);
     if (sock < 0) {
+        free(ip_host);
```
michalvasko, line 2731: Looking at the code, I could not find a path where `ip_host` would be set and `sock` `-1` so this is a false positive and redundant code.

michalvasko (changes requested):

src/session_server.c
jsc-se: OK, I will remove.

src/server_config.c
jsc-se: OK

src/server_config.c
jsc-se: OK

michalvasko (approved):

PR #559, opened 2025-10-02
session BUGFIX do not close SSH socket twice

Ilyesbenhamouda:
Currently, SSH socket is closed twice when created by libssh, causing undefined behavior in multithreaded programs (e.g. netopeer2-server).

ssh_disconnect() used to close the SSH socket. However, this behavior changed with libssh 0.10: the socket is no longer closed when passed through options.
Let's only close SSH socket when created by us after disconnecting from the session, if libssh >= 0.10 (libnetconf2 requires libssh >= 0.9.5).

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -816,12 +816,18 @@ nc_session_free_transport(struct nc_session *session, int *multisession)
                     free(siter);
                 } while (session->ti.libssh.next != session);
             }
-            /* remember sock so we can close it */
-            sock = ssh_get_fd(session->ti.libssh.session);
             if (connected) {
-                /* does not close sock */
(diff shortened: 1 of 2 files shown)


michalvasko (commented):
Why did you add the `opt_sock` member into the session instead of keeping the socket in the libssh session and getting it using `ssh_get_fd()`?

michalvasko (commented):
Actually, I think it now works correctly for `libssh` >= 0.10.0, I do remember changing this a long time ago, but 0.10.0 is 3 year old. The issue you describe is present for older `libssh` versions when the socket is really closed twice.

PR #560, opened 2025-10-03
nc_read_msg_io buf unused + don't create deleted endpoints

jeremie6wind:
Hi,
I had these 2 issues with libnetconf 4.0.1:

First patch:
src/io.c: Sometimes I get ret = 0 and an empty buf so ly_in_new_memory failed.
I propose to return if ret is 0?

Second patch:
Replacing the netconf server endpoint may fail:
For instance if the endpoint name is ssh1, we try to replace it with an endpoint name called ssh2.
nc_server_config_listen_all_endpoints first tries to create ssh1 whereas it is supposed to be deleted.
It fails and return.
Then it does not try to create ssh2.

For instance I tried to apply these 2 configs:
```
<config xmlns="urn:ietf:params:xml:ns:netconf:base:1.0">
<netconf-server xmlns="urn:ietf:params:xml:ns:yang:ietf-netconf-server">
    <listen>
        <endpoints>
            <endpoint>
                <name>ssh1</name>
                <ssh>
                    <tcp-server-parameters>
                        <local-address>127.0.0.1</local-address>
                        <local-port>830</local-port>
                    </tcp-server-parameters>
                    <ssh-server-parameters>
                      <server-identity>
                        <host-key>
                          <name>default-key</name>
                          <public-key>
                            <central-keystore-reference>genkey</central-keystore-reference>
                          </public-key>
                        </host-key>
                      </server-identity>
                    </ssh-server-parameters>
                </ssh>
            </endpoint>
        </endpoints>
    </listen>
</netconf-server>
</config>

```
then:

```
<config xmlns="urn:ietf:params:xml:ns:netconf:base:1.0">
<netconf-server xmlns="urn:ietf:params:xml:ns:yang:ietf-netconf-server">
    <listen>
        <endpoints>
            <endpoint>
                <name>ssh2</name>
                <ssh>
                    <tcp-server-parameters>
                        <local-address>127.0.0.2</local-address>
			            <local-port>830</local-port>
                    </tcp-server-parameters>
                    <ssh-server-parameters>
                      <server-identity>
                        <host-key>
                          <name>default-key</name>
                          <public-key>
                            <central-keystore-reference>genkey</central-keystore-reference>
                          </public-key>
                        </host-key>
                      </server-identity>
                    </ssh-server-parameters>
                </ssh>
            </endpoint>
        </endpoints>
    </listen>
</netconf-server>
</config>
```

I get these errors:
[ERR]: LN: Endpoint "ssh1" was not found.
[ERR]: LN: Applying ietf-netconf-server configuration failed.

In the second patch, I tried to skip endpoints with the delete yang operation.

Thank you

### DIFF
--- a/src/server_config.c
+++ b/src/server_config.c
@@ -4303,6 +4303,7 @@ nc_server_config_listen_all_endpoints(const struct lyd_node *node)
     struct lyd_node *n, *iter;
     struct nc_endpt *endpt;
     struct nc_bind *bind;
+    struct lyd_meta *m;
 
     NC_CHECK_ERR_RET(strcmp(LYD_NAME(node), "netconf-server"),
             ERR(NULL, "Unexpected node \"%s\" encountered, expected \"netconf-server\".", LYD_NAME(node)), 1);
@@ -4314,6 +4315,11 @@ nc_server_config_listen_all_endpoints(const struct lyd_node *node)
     if (n) {
         /* go over all the endpoints */
         LY_LIST_FOR(lyd_child(n), iter) {
+            if ((m = lyd_find_meta(iter->meta, NULL, "yang:operation"))) {
+                if (!strcmp(lyd_get_meta_value(m), "delete")) {
+                    continue;
+                }
+            }
             /* get the endpoint and bind structs from the node */
             NC_CHECK_RET(nc_server_config_get_endpt(iter, &endpt, &bind));
 


src/server_config.c
```diff
@@ -4314,6 +4314,13 @@ nc_server_config_listen_all_endpoints(const struct lyd_node *node)
     if (n) {
         /* go over all the endpoints */
         LY_LIST_FOR(lyd_child(n), iter) {
+            struct lyd_meta *m;
```
Roytak, line 4317: We define all variables at the beginning of the given functions.

Roytak (approved):
LGTM, just one small thing and will merge it then. Thank you.

Issue #589, opened 2026-04-06
[WRN]: LN: SSH: ssh_socket_exception_callback: Socket exception callback: 1 (0)


madis0911:
Hello,
Version Information: 
libnetconf - libnetconf2-3.7.1-1.wos1.wf42.x86_64
libssh - libssh-0.11.3-1.wf42.x86_64
netopeer2-server - netopeer2-2.4.1-1.wos1.wf42.x86_64

Problem Description
Experiencing persistent SSH socket exception callbacks when using libnetconf2 for NETCONF session management:
This error occurs consistently during session cleanup, appearing once per session that gets freed

[INF]: NP: Session 1: thread 0 event session terminated.
[INF]: NP: Generated new event (netconf-session-end).
**[WRN]: LN: SSH: ssh_socket_exception_callback: Socket exception callback: 1 (0)**
[INF]: NP: Session 2: thread 0 event new RPC.
[INF]: NP: Session 2: thread 0 event session terminated.
[INF]: NP: Generated new event (netconf-session-end).
**[WRN]: LN: SSH: ssh_socket_exception_callback: Socket exception callback: 1 (0)**
[INF]: NP: Session 3: thread 0 event new RPC.
[INF]: NP: Session 3: thread 0 event session terminated.
[INF]: NP: Generated new event (netconf-session-end).
**[WRN]: LN: SSH: ssh_socket_exception_callback: Socket exception callback: 1 (0)**
[INF]: NP: Session 4: thread 0 event new RPC.
[INF]: NP: Session 4: thread 0 event session terminated.
[INF]: NP: Generated new event (netconf-session-end).
**[[WRN]: LN: SSH: ssh_socket_exception_callback: Socket exception callback: 1 (0)]([url]([url](

attached my stub application which does below steps: 
1. Creates 4 sequential NETCONF sessions
2. Cleans up each session sequentially
3. Measures/observes socket exception behavior

[ssh_stub (1).cpp](https://github.com/user-attachments/files/26506087/ssh_stub.1.cpp)

usage: 
./ssh_stub <host> <port> <username> <password>

Please check and let me know, why and how this callback exception could be fixed.

michalvasko:
The issue should be fixed in the latest `devel`.

madis0911:
> The issue should be fixed in the latest `devel`.

Can you let me know, in which version, this fix is integrated. I can upgrade based on that.

Roytak:
[...]

PR #591, opened 2026-04-07
Avoid ssh_socket_exception_callback: Socket exception callback: 1 (0) warnings from libssh

Roytak:
Wait until EOF is received from the peer or timeout of 100ms elapses while waiting until the underlying SSH channel is closed. libssh should then update its internal state correctly and not print any warnings about normal disconnects.

Fixes https://github.com/CESNET/libnetconf2/issues/589


madis0911:
I understand based on the comments from https://github.com/CESNET/libnetconf2/issues/589, this fix is done in  libnetconf library.  
as issue 589 is still open, should I wait for the closure of this issue to get libnetconf version where fix is available?

Roytak:
> I understand based on the comments from #589, this fix is done in libnetconf library. as issue 589 is still open, should I wait for the closure of this issue to get libnetconf version where fix is available?

Yes, you should wait, but honestly it's just the warning logs and it doesn't break anything. But we found out that it is not that simple to remove the warning logs, so it might take a bit longer to finish.

michalvasko (approved):

madis0911:
in which version of libnetconf2 the above fix would be available?
is it released?

madis0911:
Can you please let me know, in which version  of libnetconf2, the above fix is available.

Roytak:
Hi, sorry I missed it. It is currently in the `devel` branch (v4.2.16) and will be released in about a month or two.

PR #594, opened 2026-04-13
Refactor CH thread management

Roytak:
Change CH threads from detached to joinable (there was a memory leak where there wasn't enough time to free the thread when the server is stopping). The server now joins all the CH threads before ending.

Also fix multiple instances of incorrectly checking the return value of `nc_rwlock_lock()`.

Added a function `nc_session_server_ch_client_dispatch_stop_if_dispatched()` that joins these CH threads on config change or server destroy.

Added a WRN log to inform the user that even though he created new CH clients, they won't be dispatched because required CBs were not set, previously this was done silently and threads of clients that were deleted weren't being stopped because of that.


src/session_server.c
```diff
@@ -3844,14 +3885,15 @@ nc_connect_ch_client_dispatch(const char *client_name, nc_server_ch_session_acqu
     NC_CHECK_SRV_INIT_RET(-1);
 
     /* CONFIG READ LOCK */
```
michalvasko: WRITE lock now.

src/session_server.h
```diff
@@ -144,8 +144,12 @@ int nc_server_init(void);
 /**
  * @brief Destroy any dynamically allocated libssh and/or libssl/libcrypto and
  *        server resources.
+ *
+ * Can be called multiple times, even if a previous call failed.
+ *
+ * @return 0 on success, 1 on error - all resources could not be freed safely.
```
michalvasko: More details what happened on `1` return, in short that it is a synchronization issue and the user can try again.

michalvasko (changes requested):

michalvasko (approved):

PR #596, opened 2026-04-16
Atomic ch flag

Roytak:
Make CH thread running flag atomic. Also fixes double free bug in `nc_server_notif_cert_exp_thread()`.


src/session_p.h
```diff
@@ -693,7 +693,7 @@ struct nc_server_ch_thread_arg {
     void *new_session_fail_cb_data;                         /**< new session fail cb data */
 
     pthread_t tid;              /**< Thread ID of the Call Home client thread. */
-    int thread_running;         /**< A boolean value that is truthy while the underlying Call Home thread is running */
+    ATOMIC_T thread_running;    /**< A boolean value that is truthy while the underlying Call Home thread is running */
```
michalvasko: I do not think `truthy` is an English word. Also, please mention that the value is atomic only so that it can be accessed without locks but is still used with a condition variable in other places.

michalvasko (changes requested):

michalvasko (approved):

PR #597, opened 2026-04-17
Correctly distinguish ssh banner and protocol string

### LINKED ISSUE #592: Banner behavior
Hi!

I was trying to set the banner with libnetconf2, but multiple line banners weren't working. Digging into the code
libnetconf2 sets the banner in `session_server_ssh.c:2040` `ssh_bind_options_set(sbind, SSH_BIND_OPTIONS_BANNER, banner)`.

It seems that it actually sets the protocol string though (defined in [RFC4253 4.2](https://datatracker.ietf.org/doc/html/rfc4253#section-4.2)). Maybe that's why only 1 line banner is supported and the length is limited to 245?
Digging through libssh, there is an `ssh_send_issue_banner()` function that would actually set the actual banner message (defined in [RFC4252 5.4](https://datatracker.ietf.org/doc/html/rfc4252#section-5.4)).

Is banner leaf representing the protocol version string is the expected behavior?

Thanks!

Roytak:
Rename previous banner to protocol string and add an API setter for it.

Add actual SSH banner and send/receive it.

NBC: renamed nc_session_ssh_get_banner API to
nc_session_ssh_get_protocol_string, to better reflect what it's actually getting

Fixes #592


src/session.c
```diff
@@ -635,20 +635,18 @@ nc_session_get_port(const struct nc_session *session)
 #ifdef NC_ENABLED_SSH_TLS
 
 API const char *
-nc_session_ssh_get_banner(const struct nc_session *session)
+nc_session_ssh_get_protocol_string(const struct nc_session *session)
```
michalvasko, line 638: To avoid NBC changes, keep the old function as well. Let it just call `nc_session_ssh_get_protocol_string()` and doxygen should say that it is deprecated and the new function should be used instead.

modules/libnetconf2-netconf-server@2025-11-11.yang
```diff
@@ -162,15 +162,18 @@ module libnetconf2-netconf-server {
...
+      type string;
+
       description
-        "The banner that will be sent to the client when connecting to the server.
-         If not set, the libnetconf2 default with its version will be used.";
+        "SSH banner sent to clients before authentication.
+         It can be used to provide information about the server or legal notices.
+         Note that the banner is sent before authentication, so it should not contain any sensitive information.
+         
+         This feature requires libssh version 0.10.0 or later. If built with an older
```
michalvasko: Such very implementation-specific information should not be in YANG. It is fine to just remove it and print a WRN message in case an old libssh version is used and the banner is set but cannot be sent.

michalvasko (changes requested):

src/session_client_ssh.c
```diff
@@ -1337,6 +1337,17 @@ connect_ssh_session(struct nc_session *session, struct nc_client_ssh_opts *opts,
         return 1;
     }
 
+#if (LIBSSH_VERSION_MAJOR > 0) || (LIBSSH_VERSION_MAJOR == 0 && LIBSSH_VERSION_MINOR >= 10)
+    char *banner;
+
+    /* retrieve the SSH banner if any was sent by the server right before authentication */
+    banner = ssh_get_issue_banner(ssh_sess);
+    if (banner) {
+        VRB(session, "Received SSH banner:\n%s", banner);
```
michalvasko: I am wondering whether `VRB` is correct verbose level here. Depending on the definition of the banner, I would expect it should always be printed and not depend on the verbose level.

michalvasko (changes requested):

michalvasko (approved):

PR #618, opened 2026-07-07
session mbedtls, openssl UPDATE enforce EKU and KU set by client

niklas-moser:
O-RAN Work Group 11 (Security Work Group) defines in Test Specification [6.9.3](https://specifications.o-ran.org/download?id=1138) the following test case:

> 
> 1. Verify a Certificate with Incorrect or Missing Usage Extensions:
> 
> - Set up a certificate with incorrect or missing key usage or extended key usage extensions.
> - Attempt to use the certificate for its intended purpose.
> - Verify that the certificate is rejected.

Here, O-RAN is stricter then RFC 5280, section [4.2.1.12](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12), which does not require the KU or EKU extensions to be set.

This PR adds the enforcement of these extensions to be set by the client (for the mbedtls and openssl implementation), thus aligning with the O-RAN specification.

Let me know what you think!
Thank you
Niklas


michalvasko:
We are implementing only the base YANG and NETCONF specification and that must be the default behavior. Supporting this O-RAN specification is fine but it must be optional so please adjust it. Whether you make it compile-dependent or add a full YANG option for TLS sessions is up to you but the latter would be preferred (much more difficult, though).

Also, please target the `devel` branch, we do not accept PRs into `master`.

niklas-moser:
Thank you! I opted for the compile-time flag for now. I will close the PR here and open it again with `devel` as target branch.

PR #619, opened 2026-07-07
session mbedtls, openssl UPDATE optionally enforce EKU and KU

niklas-moser:
Continuation of this [PR](https://github.com/CESNET/libnetconf2/pull/618), with `devel` as target branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,7 @@ option(ENABLE_DNSSEC "Enable support for SSHFP retrieval using DNSSEC for SSH (r
 option(ENABLE_PAM "Detect and use PAM" ON)
 option(ENABLE_COMMON_TARGETS "Define common custom target names such as 'doc' or 'uninstall', may cause conflicts when using add_subdirectory() to build this project" ON)
 option(ENABLE_IP_FREEBIND "Enable the IP_FREEBIND/IPV6_FREEBIND options on the listening TCP socket" OFF)
+option(ENABLE_COMPLY_WITH_ORAN_WG11 "Comply with the O-RAN WG11 security spec (e.g. enforce the clientAuth Extended Key Usage and the digitalSignature Key Usage on client certificates)" OFF)
 option(BUILD_SHARED_LIBS "By default, shared libs are enabled. Turn off for a static build." ON)
 set(READ_INACTIVE_TIMEOUT 20 CACHE STRING "Maximum number of seconds waiting for new data once some data have arrived")
 set(READ_ACTIVE_TIMEOUT 300 CACHE STRING "Maximum number of seconds for receiving a full message")
@@ -179,6 +180,11 @@ set(format_sources
 if(ENABLE_DNSSEC AND NOT ENABLE_SSH_TLS)
     message(WARNING "DNSSEC SSHFP retrieval cannot be used without SSH support.")
     set(ENABLE_DNSSEC OFF)
+endif()
+
+if(ENABLE_COMPLY_WITH_ORAN_WG11 AND NOT ENABLE_SSH_TLS)
+    message(WARNING "O-RAN WG11 compliance cannot be used without TLS support.")
+    set(ENABLE_COMPLY_WITH_ORAN_WG11 OFF)
 endif()
 
 if(ENABLE_VALGRIND_TESTS)
@@ -309,6 +315,11 @@ if(ENABLE_SSH_TLS)
 
     # set compiler flag
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DNC_ENABLED_SSH_TLS")
+
+    # comply with the O-RAN WG11 security spec (e.g. client cert Extended Key Usage and Key Usage checks)
+    if(ENABLE_COMPLY_WITH_ORAN_WG11)
+        set(NC_COMPLY_WITH_ORAN_WG11 ON)
+    endif()
 endif()
 
 # dependencies - libval
--- a/src/session_mbedtls.c
+++ b/src/session_mbedtls.c
@@ -715,6 +715,51 @@ error:
     return -1;
 }
 
+#ifdef NC_COMPLY_WITH_ORAN_WG11
+
+/**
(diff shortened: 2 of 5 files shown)


src/session_mbedtls.c
```diff
@@ -789,6 +789,37 @@ nc_server_tls_verify_cb(void *cb_data, mbedtls_x509_crt *cert, int depth, uint32
...
+            /* WG11 6.9.3 / O-RAN: the end-entity (client) certificate MUST also
+             * carry a Key Usage extension with the digitalSignature bit set (the
+             * client signs the CertificateVerify handshake message), so reject any
+             * leaf certificate missing the Key Usage extension or that bit. */
+            if (!mbedtls_x509_crt_has_ext_type(cert, MBEDTLS_X509_EXT_KEY_USAGE) ||
+                    mbedtls_x509_crt_check_key_usage(cert, MBEDTLS_X509_KU_DIGITAL_SIGNATURE)) {
+                ERR(data->session, "Cert verify: fail (client certificate lacks the required digitalSignature Key Usage).");
+                *flags |= MBEDTLS_X509_BADCERT_KEY_USAGE;
+                return 0;
+            }
```
Roytak: I think that extracting this code into a separate function would look much cleaner. If it fails, then just set the `flags` and return, also wrap the function in the `ifdef`.

src/session_openssl.c
```diff
@@ -463,6 +463,49 @@ nc_server_tls_verify_cb(int preverify_ok, X509_STORE_CTX *x509_ctx)
...
+                if (ku) {
+                    /* bit 0 of the Key Usage bit string is digitalSignature */
+                    has_digital_signature = ASN1_BIT_STRING_get_bit(ku, 0);
+                    ASN1_BIT_STRING_free(ku);
+                }
+                if (!has_digital_signature) {
+                    ERR(data->session, "Cert verify: fail (client certificate lacks the required digitalSignature Key Usage).");
+                    return 0;
+                }
+            }
```
Roytak: Do the same as for mbedTLS, we really always try to avoid inner code blocks in this codebase. Otherwise the code looks OK.

CMakeLists.txt
```diff
@@ -96,6 +96,7 @@ option(ENABLE_DNSSEC "Enable support for SSHFP retrieval using DNSSEC for SSH (r
 option(ENABLE_PAM "Detect and use PAM" ON)
 option(ENABLE_COMMON_TARGETS "Define common custom target names such as 'doc' or 'uninstall', may cause conflicts when using add_subdirectory() to build this project" ON)
 option(ENABLE_IP_FREEBIND "Enable the IP_FREEBIND/IPV6_FREEBIND options on the listening TCP socket" OFF)
+option(ENABLE_COMPLY_WITH_ORAN_WG11 "Comply with the O-RAN WG11 security spec (e.g. enforce the clientAuth Extended Key Usage and the digitalSignature Key Usage on client certificates), non-default behavior" OFF)
```
Roytak: I think you can omit the "non-default behavior" here.

Roytak (changes requested):
Looks good, but needs a bit of refactor.

niklas-moser:
Thank you! I updated the branch, let me know what you think

Roytak (approved):

CMakeLists.txt
```diff
@@ -309,6 +315,11 @@ if(ENABLE_SSH_TLS)
 
     # set compiler flag
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DNC_ENABLED_SSH_TLS")
+
+    # comply with the O-RAN WG11 security spec (e.g. client cert Extended Key Usage and Key Usage checks)
+    if(ENABLE_COMPLY_WITH_ORAN_WG11)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DCOMPLY_WITH_ORAN_WG11")
```
Roytak: Prepend `NC_` to the compile def name, so it's `NC_COMPLY_WITH_ORAN_WG11`, similarly to `NC_ENABLED_SSH_TLS`.

Roytak (commented):
Last comment from me, otherwise LGTM. Will wait for Michal to look at it and merge it when he comes back next week.

CMakeLists.txt
niklas-moser: Done!

niklas-moser:
Okay, updated the PR. Thanks!

CMakeLists.txt
```diff
@@ -309,6 +315,11 @@ if(ENABLE_SSH_TLS)
 
     # set compiler flag
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DNC_ENABLED_SSH_TLS")
+
+    # comply with the O-RAN WG11 security spec (e.g. client cert Extended Key Usage and Key Usage checks)
+    if(ENABLE_COMPLY_WITH_ORAN_WG11)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DNC_COMPLY_WITH_ORAN_WG11")
+    endif()
```
michalvasko, line 322: Please replace this block with `#cmakedefine ENABLE_COMPLY_WITH_ORAN_WG11` in `config.h.in`.

michalvasko (changes requested):
Just some minor changes are needed. As for formatting (the failed tests), a newline is expected before `#endif`. But please also add one after every `#ifxxx`, thanks.

niklas-moser:
Thank you, I updated the PR with the newlines and using `#cmakedefine` through `config.h.in`.

michalvasko (approved):

michalvasko (changes requested):
I am not exactly sure why, but `uncrustify` does not like the newlines in `session_mbedtls.c` on `838` and 843` and `session_openssl.c` on `519` and `524`. Best to just run `uncrustify` locally but if you remove these 4 newlines, the tests should finally pass.

niklas-moser:
Done!

michalvasko (approved):

PR #621, opened 2026-07-16
add max-auth-attempts

vxpyy9:
Limit the number of failed SSH authentication attempts before the server disconnects the client, default 3.

### DIFF
--- a/modules/libnetconf2-netconf-server@2026-07-16.yang
+++ b/modules/libnetconf2-netconf-server@2026-07-16.yang
@@ -31,6 +31,10 @@ module libnetconf2-netconf-server {
     prefix tlss;
   }
 
+  revision "2026-07-16" {
+    description "Added max-auth-attempts leaf to limit SSH authentication attempts.";
+  }
+
   revision "2026-04-17" {
     description "Change SSH banner description, reference and string length to reflect its correct purpose.";
   }
@@ -158,6 +162,15 @@ module libnetconf2-netconf-server {
       units "seconds";
       description
         "Represents the maximum amount of seconds an authentication can go on for.";
+    }
+
+    leaf max-auth-attempts {
+      type uint16 {
+        range "1..max";
+      }
+      default 3;
+      description
+        "Maximum number of failed SSH authentication attempts before disconnecting the client.";
     }
   }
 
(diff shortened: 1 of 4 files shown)


modules/libnetconf2-netconf-server@2026-07-16.yang
```diff
@@ -159,6 +163,15 @@ module libnetconf2-netconf-server {
...
     }
+
+    leaf max-auth-attempts {
+      type uint16 {
+        range "1..max";
+      }
+      default 3;
+      description
+        "Maximum number of failed SSH authentication attempts before disconnecting the client.";
+    }
```
Roytak, line 174: Mention in the description that it is per connection. I think that the [RFC4252](https://datatracker.ietf.org/doc/html/rfc4252) should be referenced here.

Furthermore, this description should probably mention something along the lines "each failed authentication request, including rejected public keys, counts as one attempt", because some SSH clients try all their configured public keys and hence hitting the 3 might be easy, so just so the users are aware.

src/session_server_ssh.c
Roytak, line 1645: ssh_auth_attempts need to be increased here, otherwise it would always remain 0 and that would lead to unlimited authenticaiton attempts (e.g. 3 would never be reached by the calling function) + we need to either reach the code checking if we hit max attempts from here or copy it here.

src/session_server_ssh.c
```diff
@@ -1727,7 +1727,13 @@ nc_server_ssh_auth(struct nc_session *session, struct nc_server_ssh_opts *opts,
         ++session->opts.server.ssh_auth_attempts;
         VRB(session, "Failed user \"%s\" authentication attempt (#%d).", session->username,
                 session->opts.server.ssh_auth_attempts);
-        ssh_message_reply_default(msg);
+        if (session->opts.server.ssh_max_auth_attempts &&
+                (session->opts.server.ssh_auth_attempts >= session->opts.server.ssh_max_auth_attempts)) {
+            /* disconnect immediately so client does not prompt again */
```
Roytak, line 1732: I think that both ssh_message_reply_default and ssh_disconnect should be called here, the abrupt disconnect is unnecessary.

src/session_server_ssh.c
Roytak, line 1707: We could make this more consistent, by just setting `ret = 1` and then down below the max attempts check would happen + the possible disconnect. Mention this in a comment above the VRB.

Roytak (changes requested):
Hi, thank you for the contribution. This feature was part of the code a couple years ago, but we decided to remove it. Is there any specific reason why you find it useful to have? I found a couple things that will need to be adjusted and also you should add a new test e.g. to `test_ssh.c` that tests this.

PR #622, opened 2026-07-23
Migrate to callback-based authentication for libssh 0.12

HanzlikPetr:


src/session.c
```diff
@@ -982,6 +985,11 @@ nc_session_free_transport(struct nc_session *session, int *multisession)
             sock = -1;
 #endif
 
+            /* free heap-allocated callback data (libssh >= 0.12) */
+#if LIBSSH_0_12
```
michalvasko, line 1001: `#if` should not really be needed, `cb_data` will simply be `NULL` otherwise and the `free` function will then immediately return.

src/session.c
```diff
@@ -994,6 +1002,13 @@ nc_session_free_transport(struct nc_session *session, int *multisession)
                 /* there are still multiple sessions, keep the ring list */
                 siter->ti.libssh.next = session->ti.libssh.next;
             }
+            /* transfer cb_data to a remaining session so it's freed when the SSH session is freed */
+#if LIBSSH_0_12
```
michalvasko, line 1029: Again, should be safe to simply remove `#if`.

src/session_p.h
```diff
@@ -227,8 +227,6 @@ enum nc_store_type {
 #ifdef NC_ENABLED_SSH_TLS
 
 #include <curl/curl.h>
-#include <libssh/libssh.h>
```
michalvasko, line 224: Why was it deleted? There are libssh-specific structures in the internal structures defined in this file that need their definition.

src/session_server_ssh.c
```diff
@@ -2008,15 +1180,50 @@ static int
...
 
     DBG(session, "SSH authentication...");
 
     /* authenticate */
     if (opts->auth_timeout) {
         nc_timeouttime_get(&ts_timeout, opts->auth_timeout * 1000);
     }
+#if LIBSSH_0_12
+    /* Create an event loop */
+    event = ssh_event_new();
```
michalvasko, line 1656: Would it make sense to store this structure in the session avoiding the need to always create it?

michalvasko (changes requested):

src/session_server_ssh.c
```diff
@@ -2008,15 +1172,48 @@ static int
...
+        if (!ssh_is_connected(session->ti.libssh.session)) {
+            ERR(session, "Communication SSH socket unexpectedly closed.");
+            return -1;
+        }
+
+        time_diff = nc_timeouttime_cur_diff(&ts_timeout);
+        if (time_diff < 1) {
+            /* timeout */
+            break;
+        }
```
Roytak: When `auth_timeout` is 0, `ts_timeout` is uninitialized and
`nc_timeouttime_cur_diff()` reads garbage. The `#if LIBSSH_0_12` path needs
the same `if (opts->auth_timeout && ...)` guard as the `#else` path, or
`ts_timeout` should always be initialized.

src/session.c
```diff
@@ -982,6 +985,12 @@ nc_session_free_transport(struct nc_session *session, int *multisession)
             sock = -1;
 #endif
 
+            /* free heap-allocated callback data and ssh event (libssh >= 0.12) */
+#if LIBSSH_0_12
+            nc_server_ssh_cb_data_free(session->ti.libssh.cb_data);
+            ssh_event_free(session->ti.libssh.event);
```
Roytak: Move `nc_server_ssh_cb_data_free()` to **after** `ssh_free()`
so the callback structures remain valid for the entire session lifetime as [libssh](https://api.libssh.org/stable/group__libssh__callbacks.html#gaa321f33172cfa54eb50baaa963457a17) requires.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+    n_answers = ssh_userauth_kbdint_getnanswers(session->ti.libssh.session);
+    if (n_answers < 0) {
+        ERR(session, "Failed to get number of kbdint answers.");
+        nc_server_ssh_cb_pam_cancel(pam_data);
+        cb_data->pam_kbdint = NULL;
+        return SSH_AUTH_DENIED;
+    }
+
+    answers = calloc(n_answers, sizeof *answers);
+    NC_CHECK_ERRMEM_GOTO(!answers, rc = SSH_AUTH_DENIED, cleanup);
```
Roytak: Guard the cleanup loop with `if (answers)`, or set `n_answers = 0` before jumping to cleanup when `calloc` fails. Otherwise, the cleanup loop then dereferences `answers[i]` (i.e. `NULL[i]`), which is UB.

src/session_server.c
```diff
@@ -42,6 +42,7 @@
 #include "session_p.h"
 #include "session_server.h"
 #include "session_server_ch.h"
+#include "session_server_ssh_wrapper.h"
```
Roytak: I think this probably doesn't compile with `-DENABLE_SSH_TLS=OFF`. Move the `#include "session_server_ssh_wrapper.h"` inside the `#ifdef NC_ENABLED_SSH_TLS` block, or guard the header's libssh includes with `#ifdef NC_ENABLED_SSH_TLS`.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+    /* Extract the username from the message. */
+    if (ssh_message_auth_kbdint_is_response(message) && session->username) {
+        user = session->username;
+    } else {
+        user = ssh_message_auth_user(message);
+    }
+
+    if (nc_server_ssh_cb_auth_common_setup(cb_data, user, &local_users_supported, &auth_client)) {
+        return SSH_AUTH_DENIED;
+    }
```
Roytak, line 671: Cancel any in-progress PAM exchange before returning `SSH_AUTH_DENIED` from the kbdint callback when `cb_data->pam_kbdint` is set. Otherwise, on the next kbdint request,
`nc_server_ssh_cb_kbdint_pam_request` overwrites `cb_data->pam_kbdint`, leaking the old `pam_data` and its thread.

src/session_server_ssh.c
```diff
@@ -1293,647 +1052,52 @@ nc_server_ssh_send_banner(struct nc_session *session, struct nc_server_ssh_opts
...
-        case SSH_GLOBAL_REQUEST_CANCEL_TCPIP_FORWARD:
-            str_subtype = "cancel-tcpip-forward";
-            break;
-        case SSH_GLOBAL_REQUEST_UNKNOWN:
-        /* fallthrough */
-        default:
-            str_subtype = "unknown";
+        /* This function listens to the network and automatically calls callback funcitons. */
+        ret = ssh_event_dopoll(session->ti.libssh.event, time_diff);
+        if ((ret == SSH_ERROR) || (ret == SSH_AGAIN)) {
```
Roytak: Treat `SSH_ERROR` as an error (return -1), not a timeout (return 0). At minimum, log the error before breaking.

src/session_server_ssh.c
```diff
@@ -2008,15 +1172,48 @@ static int
...
+        if (time_diff < 1) {
+            /* timeout */
+            break;
+        }
+
+        /* This function listens to the network and automatically calls callback funcitons. */
+        ret = ssh_event_dopoll(event, time_diff);
+        if ((ret == SSH_ERROR) || (ret == SSH_AGAIN)) {
+            break;
+        }
```
Roytak, line 1695: Treat `SSH_ERROR` as an error (return -1), not a timeout (return 0). At minimum, log the error before breaking.

src/session_server_ssh.c
```diff
@@ -2190,6 +1410,13 @@ nc_accept_ssh_session(struct nc_session *session, struct nc_server_ssh_opts *opt
     rc = nc_accept_ssh_session_auth(session, opts);
     session->data = NULL;
     if (rc != 1) {
+#if LIBSSH_0_12 && defined (HAVE_LIBPAM)
+        /* if PAM thread is still running, cancel and clean it up */
+        if (cb_data) {
+            nc_server_ssh_cb_pam_cancel(cb_data->pam_kbdint);
+            cb_data->pam_kbdint = NULL;
+        }
+#endif
```
Roytak, line 1900: Add a PAM cancel call on the success path too (or in `nc_server_ssh_cb_data_free` as a safety net), e.g. after `rc == 1` or before the cleanup label, to ensure no PAM thread is left running when auth succeeds via a non-kbdint method. Note that this may only happen in the "non local users" configuration.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+        pthread_mutex_unlock(&data->lock);
+        free(prompts);
+        free(echo);
+        free(*resp);
+        *resp = NULL;
+        return PAM_CONV_ERR;
+    }
+
+    /* state == NC_PAM_ANSWERS_READY: copy answers */
+    if (data->n_answers > n_requests) {
```
Roytak: Also reject `n_answers < n_requests` (use `!=` like the old code), or document why fewer answers are now acceptable.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+ * @param[in] message SSH message for sending additional prompts if needed.
+ * @return SSH_AUTH_INFO if more prompts needed.
+ * @return SSH_AUTH_SUCCESS/SSH_AUTH_DENIED on completion.
+ */
+static int
+nc_server_ssh_cb_kbdint_pam_response(struct nc_server_ssh_cb_data *cb_data, ssh_message message)
+{
+    struct nc_server_ssh_cb_pam_data *pam_data = cb_data->pam_kbdint;
+    struct nc_session *session = cb_data->session;
+    int n_answers, i, rc, n_prompts;
```
Roytak: Initialize `rc` (e.g. to `SSH_AUTH_DENIED`).

src/session_server_ssh.c
```diff
@@ -45,8 +45,108 @@
...
+{
+    if (auth_state->method_count) {
+        return;
+    }
+
+    if (local_users_supported) {
+        if (auth_client->pubkey_store != NC_STORE_UNKNOWN) {
+            auth_state->methods |= SSH_AUTH_METHOD_PUBLICKEY;
+            auth_state->method_count++;
+        }
```
Roytak, line 114: I see the condition changed from whitelist to blacklist, which makes more sense, but now also possibly accepts `NC_STORE_KEYSTORE`. Just verify this is unreachable with that value.

src/session_server_ssh.c
```diff
@@ -45,8 +45,108 @@
...
+    for (uint32_t u = 0; u < LY_ARRAY_COUNT(opts->auth_clients); u++) {
+        if (!strcmp(opts->auth_clients[u].username, user)) {
+            return &opts->auth_clients[u];
+        }
+    }
+
+    /* client not known by the endpt, but it references another one so try it */
+    if (opts->referenced_endpt_name) {
+        if (nc_server_endpt_get(opts->referenced_endpt_name, &referenced_endpt)) {
+            return NULL;
```
Roytak, line 95: Log an error (ERRINT or ERR) when the referenced endpoint cannot be found, so configuration errors are not silently ignored.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+
+    /* handle the result from the kbdint system dispatch */
+    if (ret == SSH_AUTH_INFO) {
+        /* prompts sent, waiting for client response — libssh sends no reply */
+        return SSH_AUTH_INFO;
+    } else if (ret == SSH_AUTH_SUCCESS) {
+        VRB(session, "User \"%s\" authenticated via keyboard-interactive.", user);
+        return nc_ssh_auth_success(session, &cb_data->auth_state, SSH_AUTH_METHOD_INTERACTIVE);
+    } else {
+        VRB(session, "User \"%s\" authentication denied via keyboard-interactive.", user);
```
Roytak, line 702: Add the "authenticated" and "failed authentication attempt"
log messages to the password, pubkey, and none auth callbacks for parity with the old behavior.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+            free(pubkeys[i].name);
+            free(pubkeys[i].data);
+        }
+        free(pubkeys);
+    }
+
+    if (ret == 0) {
+        if (signature_state == SSH_PUBLICKEY_STATE_NONE) {
+            /* just checking if the public key would be accepted */
+            return SSH_AUTH_SUCCESS;
```
Roytak, line 597: Even though this was missing in the old code, based on the libssh [docs](https://api.libssh.org/stable/group__libssh__callbacks.html#ga68f74aa25d6a193c5fcf016d6d1bdc51) you should also explicitly check for `SSH_PUBLICKEY_STATE_VALID` and return `SSH_AUTH_DENIED` for any other unexpected value.

src/session_server_ssh.c
```diff
@@ -45,8 +45,108 @@
...
+    if (!mod) {
+        ERRINT;
+        return -1;
+    }
+
+    if (lys_feature_value(mod, "local-users-supported") == 0) {
+        return 1;
+    }
+
+    return 0;
```
Roytak: Check for `LY_SUCCESS` (supported), `LY_ENOTFOUND` (not
supported), and treat everything else as fatal (`return -1`).

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+            session->opts.server.ssh_auth_attempts++;
+            return SSH_AUTH_DENIED;
+        }
+    }
+
+    /* handle the result from the kbdint system dispatch */
+    if (ret == SSH_AUTH_INFO) {
+        /* prompts sent, waiting for client response — libssh sends no reply */
+        return SSH_AUTH_INFO;
+    } else if (ret == SSH_AUTH_SUCCESS) {
```
Roytak: Either explicitly handle `SSH_AUTH_PARTIAL` or document that the custom callback must only return 0 or non-zero (not
`SSH_AUTH_PARTIAL`). Based on libssh docs, `SSH_AUTH_PARTIAL` is a valid return.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+                if (*p == channel_data) {
+                    *p = channel_data->next;
+                    channel_data->next = NULL;
+                    break;
+                }
+                p = &(*p)->next;
+            }
+        }
+        new_session->ti.libssh.channel_cb = &channel_data->channel_cb;
+    }
```
Roytak: Move the `p` declaration to the top of the function or restructure the search into a small helper function. Also adjust the comment into a single block or reword it.

src/session_server_ssh.c
```diff
@@ -2073,6 +1275,24 @@ nc_accept_ssh_session(struct nc_session *session, struct nc_server_ssh_opts *opt
         goto cleanup;
     }
 
+#if LIBSSH_0_12
+    cb_data = calloc(1, sizeof(*cb_data));
```
Roytak: No parentheses in sizeof.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+
+    /* create the new channel */
+    chan = ssh_channel_new(libssh_sess);
+    if (!chan) {
+        ERR(session, "Session %u: failed to create a new SSH channel.", session->id);
+        return NULL;
+    }
+
+    channel_data = calloc(1, sizeof(*channel_data));
```
Roytak: No parentheses in sizeof.

src/session_server_ssh.c
```diff
@@ -1293,647 +1052,52 @@ nc_server_ssh_send_banner(struct nc_session *session, struct nc_server_ssh_opts
...
-            str_subtype = "tcpip-forward";
-            break;
-        case SSH_GLOBAL_REQUEST_CANCEL_TCPIP_FORWARD:
-            str_subtype = "cancel-tcpip-forward";
-            break;
-        case SSH_GLOBAL_REQUEST_UNKNOWN:
-        /* fallthrough */
-        default:
-            str_subtype = "unknown";
+        /* This function listens to the network and automatically calls callback funcitons. */
```
Roytak: Typo in functions.

src/session_server_ssh.c
```diff
@@ -2008,15 +1172,48 @@ static int
...
+            return -1;
+        }
+
+        time_diff = nc_timeouttime_cur_diff(&ts_timeout);
+        if (time_diff < 1) {
+            /* timeout */
+            break;
+        }
+
+        /* This function listens to the network and automatically calls callback funcitons. */
```
Roytak: Typo in functions.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+    /* Get the stored password */
+    if (local_users_supported) {
+        stored_password = auth_client->password;
+        if (!stored_password) {
+            DBG(session, "User \"%s\" does not have password method configured.", user);
+            session->opts.server.ssh_auth_attempts++;
+            return SSH_AUTH_DENIED;
+        }
+    } else {
+    #ifdef HAVE_SHADOW
```
Roytak: Move the ifdef to column 0.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+{
+    struct nc_server_ssh_cb_data *cb_data = (struct nc_server_ssh_cb_data *)userdata;
+    struct nc_session *session = cb_data->session;
+    struct nc_auth_client *auth_client = NULL;
+    int local_users_supported = 0;
+
+    if (nc_server_ssh_cb_auth_common_setup(cb_data, user, &local_users_supported, &auth_client)) {
+        if (local_users_supported && !auth_client) {
+            ssh_set_auth_methods(session->ti.libssh.session, SSH_AUTH_METHOD_PUBLICKEY);
+        }
```
Roytak: Remove the redundant `ssh_set_auth_methods` call, or add a
comment explaining why it's needed in `nc_server_ssh_cb_auth_none` but not in the other auth callbacks. It is already called in `nc_server_ssh_cb_auth_common_setup`.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+                echo[j] = 1;
+            }
+            j++;
+        }
+    }
+
+    /* signal main thread that prompts are ready */
+    pthread_mutex_lock(&data->lock);
+    if (data->state == NC_PAM_CANCELLED) {
+        pthread_mutex_unlock(&data->lock);
```
Roytak: Add `/* LOCK */` / `/* UNLOCK */` comments around all the `pthread_mutex_lock`/`pthread_mutex_unlock` calls in the PAM bridge code.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+    /* read answers from libssh kbdint structure */
+    n_answers = ssh_userauth_kbdint_getnanswers(session->ti.libssh.session);
+    if (n_answers < 0) {
+        ERR(session, "Failed to get number of kbdint answers.");
+        nc_server_ssh_cb_pam_cancel(pam_data);
+        cb_data->pam_kbdint = NULL;
+        return SSH_AUTH_DENIED;
+    }
+
+    answers = calloc(n_answers, sizeof *answers);
```
Roytak: If `n_answers` is 0 (client sent 0 answers), `calloc(0, sizeof *answers)` may return NULL, which triggers NC_CHECK_ERRMEM_GOTO` as a false OOM error.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+ * @brief Phase 1 of callback-based PAM kbdint: start PAM thread, wait for prompts, send to client.
+ *
+ * @param[in] cb_data Callback data (stores pam_kbdint pointer for Phase 2).
+ * @param[in] message SSH message for sending prompts.
+ * @param[in] user Username.
+ * @return SSH_AUTH_INFO if prompts sent.
+ * @return SSH_AUTH_SUCCESS/SSH_AUTH_DENIED on PAM completion.
+ */
+static int
+nc_server_ssh_cb_kbdint_pam_request(struct nc_server_ssh_cb_data *cb_data, ssh_message message, const char *UNUSED(user))
```
Roytak: Either use the `user` parameter or remove it from the function signature (and adjust the caller).

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+            while (*p) {
+                if (*p == channel_data) {
+                    *p = channel_data->next;
+                    channel_data->next = NULL;
+                    break;
+                }
+                p = &(*p)->next;
+            }
+        }
+        new_session->ti.libssh.channel_cb = &channel_data->channel_cb;
```
Roytak: If `channel_data` is not found in the list (shouldn't happen, but no assert), the code silently continues and assigns `channel_cb` to the new session anyway, potentially leading to a double-free. Add an `assert(*p == channel_data)` or error handling after the while loop.

src/session_server.c
```diff
@@ -2294,6 +2295,34 @@ nc_server_send_reply_io(struct nc_session *session, int io_timeout, const struct
...
 
+#ifdef NC_ENABLED_SSH_TLS
+/**
+ * @brief Scan the session ring for a newly established NETCONF SSH channel.
+ *
+ * @param[in] session Session whose SSH channel ring to scan.
+ * @return Pointer to the new starting NETCONF session if found, NULL otherwise.
+ */
+static struct nc_session *
+nc_ps_ssh_find_new_channel(struct nc_session *session)
```
Roytak: Consider returning `int` (1/0) from `nc_ps_ssh_find_new_channel` since the pointer is never used by callers.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+ *     https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#define _GNU_SOURCE
+
+#include "config.h" /* Expose HAVE_LIBPAM and HAVE_SHADOW */
+
+#ifdef HAVE_LIBPAM
+#   include <security/pam_appl.h>
+#endif
```
Roytak: Move under system headers.

src/session_server_ssh_wrapper.h
```diff
@@ -0,0 +1,316 @@
...
+#define _SESSION_SERVER_SSH_WRAPPER_H_
+
+#include "config.h" /* Expose HAVE_LIBPAM, HAVE_SHADOW */
+
+#ifdef HAVE_LIBPAM
+#   include <security/pam_appl.h>
+#endif
+#ifdef HAVE_SHADOW
+#   include <shadow.h>
+#endif
```
Roytak, line 37: Move config.h above session_p.h and the rest below.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+        }
+        session->ti.libssh.next = new_session;
+
+        new_session->status = NC_STATUS_STARTING;
+        new_session->ti_type = NC_TI_SSH;
+        new_session->io_lock = session->io_lock;
+        new_session->ti.libssh.channel = channel;
+        new_session->ti.libssh.session = session->ti.libssh.session;
+        new_session->username = strdup(session->username);
+        new_session->host = strdup(session->host);
```
Roytak: Add `NC_CHECK_ERRMEM_GOTO` checks for the `strdup` calls.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+        /* PAM thread is waiting for answers — tell it to abort */
+        data->state = NC_PAM_CANCELLED;
+        pthread_cond_signal(&data->changed);
+    } else if (data->state == NC_PAM_RUNNING) {
+        /* PAM thread is processing — mark as cancelled so it stops at next conv */
+        data->state = NC_PAM_CANCELLED;
+    }
+    pthread_mutex_unlock(&data->lock);
+
+    pthread_join(data->thread, NULL);
```
Roytak, line 197: If a PAM module block indefinitely, this call would also block. Consider using `pthread_timedjoin_np` with the remaining `auth_timeout` or another mechanism.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+    /* signal PAM thread to cancel */
+    pthread_mutex_lock(&data->lock);
+    if (data->state == NC_PAM_PROMPTS_READY) {
+        /* PAM thread is waiting for answers — tell it to abort */
+        data->state = NC_PAM_CANCELLED;
+        pthread_cond_signal(&data->changed);
+    } else if (data->state == NC_PAM_RUNNING) {
+        /* PAM thread is processing — mark as cancelled so it stops at next conv */
+        data->state = NC_PAM_CANCELLED;
+    }
```
Roytak: This feels fragile, I think you should instead add a defensive `else` branch that sets `NC_PAM_CANCELLED` for any state other than `NC_PAM_DONE`/`NC_PAM_CANCELLED`.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,1016 @@
...
+    int ret;
+
+    conv.conv = nc_server_ssh_cb_pam_conv;
+    conv.appdata_ptr = data;
+
+    if (!server_opts.pam_config_name) {
+        ERR(data->session, "PAM configuration filename not set.");
+        ret = 1;
+        goto done;
+    }
```
Roytak: Why not check this and other pre-conditions, if any, before creating the thread?

Roytak (changes requested):
Looks okay, most of the comments were small things. I also noticed there is now significant code duplication between the callback and message paths (e.g. shadow kbdint, pubkey comparison and such), these could probably be moved to `session_server_ssh.c` and shared between the two.

src/session_server_ssh.c
HanzlikPetr: Verified: unreachable. `pubkey_store` is only assigned in `server_config.c` and can only be `LOCAL` / `TRUSTSTORE` / `SYSTEM` / `UNKNOWN` (a keystore ref exists in the model only for server private keys and TLS certs, not user public keys). No assignment of `NC_STORE_KEYSTORE` to this field exists anywhere in the tree, and the field is zero-init to `NC_STORE_UNKNOWN` by default.

src/session.c
```diff
@@ -994,6 +1007,15 @@ nc_session_free_transport(struct nc_session *session, int *multisession)
...
             }
+            /* transfer cb_data to a remaining session so it's freed when the SSH session is freed */
+            if (session->ti.libssh.cb_data) {
+                siter->ti.libssh.cb_data = session->ti.libssh.cb_data;
+                session->ti.libssh.cb_data = NULL;
+            }
+            if (session->ti.libssh.event) {
+                siter->ti.libssh.event = session->ti.libssh.event;
+                session->ti.libssh.event = NULL;
+            }
```
Roytak, line 1040: The transfer moves `cb_data` and `event` to a surviving ring session, but `cb_data->session` keeps pointing at the session that is being freed, enabling a UAF.

src/session_server_ssh.c
```diff
@@ -40,1900 +33,1520 @@
...
+    } else {
+        new_session->ti.libssh.next = session->ti.libssh.next;
+    }
+    session->ti.libssh.next = new_session;
+
+    return new_session;
+
+error:
+    nc_session_free(new_session, NULL);
+    return NULL;
```
Roytak, line 547: This error path runs while `new_session` still shares the parent's `ti.libssh.session` and `io_lock` and is not in the ring yet (`next == NULL`). `nc_session_free()` therefore takes the sole owner branch
and also frees shared members. The cleanup added for the strdup checks should
detach the shared state first.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,846 @@
...
+ */
+static int
+nc_server_ssh_cb_auth_common_setup(struct nc_server_ssh_cb_data *cb_data, const char *user,
+        int *local_users_supported, struct nc_auth_client **auth_client)
+{
+    struct nc_session *session = cb_data->session;
+    struct nc_server_ssh_opts *opts = cb_data->opts;
+
+    *local_users_supported = 0;
+    *auth_client = NULL;
```
Roytak, line 485: The message path refuses userauth requests received after the session is
authenticated (`nc_session_ssh_msg()`:
`if (session->flags & NC_SESSION_SSH_AUTHENTICATED) { ... reply_default }`),
but none of the callback entry points has an equivalent check, and libssh
does not filter post-auth userauth requests itself.

Please add the parity check early in `nc_server_ssh_cb_auth_common_setup()`:

    if (session->flags & NC_SESSION_SSH_AUTHENTICATED) {
        ERR(session, "User \"%s\" authenticated, but requested another authentication.", session->username);
        return -1;
    }

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,846 @@
...
+    }
+
+    assert(*chan_ptr == channel_data);
+    if (*chan_ptr) {
+        *chan_ptr = channel_data->next;
+        channel_data->next = NULL;
+    }
+
+    new_session->ti.libssh.channel_cb = &channel_data->channel_cb;
+    return 0;
```
Roytak: If a client sends a second "netconf" subsystem request on the same additional channel, `channel_data` is no longer in `cb_data->channels` (it was unlinked when the first request claimed it). The `assert(*chan_ptr == channel_data)` then aborts Debug builds on client input, and in Release builds eventually leading to double-free.

Please handle the lookup failure explicitly, e.g. `if (!*chan_ptr) return 1;` (channel already claimed) , dropping the assert. Hardening `nc_server_ssh_channel_subsys_check()` to also scan the ring for an existing owner of `channel` would make the duplicate request fail earlier and more clearly.

src/session_server_ssh.c
```diff
@@ -40,1900 +33,1520 @@
...
     }
 
-    fclose(file);
-
-    /* checksum */
-    if (privkey_format) {
-        len = 10 + strlen(privkey_format) + 17 + strlen(in) + 9 + strlen(privkey_format) + 16;
-    } else {
-        len = 10 + 18 + strlen(in) + 9 + 17;
+    for (uint32_t u = 0; u < LY_ARRAY_COUNT(opts->auth_clients); u++) {
```
Roytak: Move var declaration to the beginning.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,846 @@
...
+    }
+
+    /* create the new channel */
+    chan = ssh_channel_new(libssh_sess);
+    if (!chan) {
+        ERR(session, "Session %u: failed to create a new SSH channel.", session->id);
+        return NULL;
+    }
+
+    channel_data = calloc(1, sizeof(*channel_data));
```
Roytak: Change to `sizeof *channel_data` per coding style.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,846 @@
...
+    if (*local_users_supported) {
+        *auth_client = nc_ssh_find_auth_client(opts, user, session);
+
+        if (!*auth_client) {
+            ERR(session, "User \"%s\" not known by the server.", user);
+            /* advertise only publickey so there is no interaction and it is simply denied */
+            ssh_set_auth_methods(session->ti.libssh.session, SSH_AUTH_METHOD_PUBLICKEY);
+            nc_server_ssh_auth_attempt_failed(session);
+            return -1;
+        }
```
Roytak, line 526: `nc_server_ssh_auth_common_setup()` counts/logs a failed auth attempt for unknown users (and other early setup failures) on every method attempt, while the message path (`nc_server_ssh_msg_auth()`) replies to unknown users without touching `ssh_auth_attempts`. So the two builds now behave differently for the same traffic. Up to you which behavior is correct, but please make both paths consistent (counting unknown-user attempts seems the more useful policy).

src/session.c
```diff
@@ -965,6 +967,7 @@ nc_session_free_transport(struct nc_session *session, int *multisession)
                     /* free starting SSH NETCONF session (channel will be freed in ssh_free()) */
                     free(siter->username);
                     free(siter->host);
+                    free(siter->ti.libssh.channel_cb);
```
Roytak: Same class of lifetime-contract issue as the one fixed for `cb_data` below: this frees the channel callback structs while the siblings' channels are still alive in the libssh session (they die only in `ssh_free()` further down). libssh requires channel callbacks to remain valid for the whole channel lifetime. This seems currently unreachable, but collecting the pointers and freeing them after `ssh_free()` would make it robust.

src/session_server_ssh_wrapper.h
```diff
@@ -0,0 +1,471 @@
...
+#endif /* HAVE_LIBPAM */
+
+#ifdef HAVE_SHADOW
+/**
+ * @brief Get the user's hashed password from the system.
+ *
+ * @param[in] username Username.
+ * @return User's hashed password or NULL on error.
+ */
+char * nc_server_ssh_get_pwd_hash(const char *username);
```
Roytak: Should be `char *nc_server_ssh_get_pwd_hash(const char *username);`

src/session_server_ssh.c
```diff
@@ -40,1900 +33,1520 @@
...
+        }
     }
 
-    if ((unsigned)written != len) {
-        unlink(path);
-        return NULL;
+    /* client not known by the endpt, but it references another one so try it */
+    if (opts->referenced_endpt_name) {
+        if (nc_server_endpt_get(opts->referenced_endpt_name, &referenced_endpt)) {
+            ERR(session, "Referenced endpoint \"%s\" not found.", opts->referenced_endpt_name);
```
Roytak, line 94: `nc_server_endpt_get()` already logs an error when the endpoint is not found, so this produces two error messages per failed lookup. Please keep just one of them, I would prefer `nc_server_endpt_get` not to log anything and to leave it to the callers, but that would most likely require adding this `ERR` log to all the callers.

src/session_server_ssh.c
```diff
@@ -2007,16 +1620,58 @@ nc_ssh_bind_add_hostkeys(ssh_bind sbind, struct nc_server_ssh_opts *opts)
...
     DBG(session, "SSH authentication...");
 
     /* authenticate */
     if (opts->auth_timeout) {
         nc_timeouttime_get(&ts_timeout, opts->auth_timeout * 1000);
     }
+#if LIBSSH_0_12
+    /* Create an event loop */
+    event = ssh_event_new();
+    ssh_event_add_session(event, session->ti.libssh.session);
```
Roytak: Check return values.

src/session_server_ssh_auth_callback.c
```diff
@@ -0,0 +1,846 @@
...
+            ERR(data->session, "Unable to cancel a PAM thread, leaking its resources.");
+            return;
+        }
+    } else {
+        pthread_join(data->thread, NULL);
+    }
+#else
+    (void)cancelled;
+    pthread_join(data->thread, NULL);
+#endif
```
Roytak, line 198: I suggest using the timed join for all paths (when the macro is available), which also removes the `cancelled` variable and makes it much simpler overall. The way I imagine this is to: lock the mutex, set state to `NC_PAM_CANCELLED` and for the cases where it makes sense also call `pthread_cond_signal`/`pthread_cancel` (should be safe to unlock the mutex after cancel), afterwards just call `pthread_timedjoin_np` and if it fails, just accept the leak - this should work, unless I am missing something.

Roytak (changes requested):
Reviewed. We should also eventually add more tests for PAM, since only the 'happy' path is tested now, without any timeouts/cancels/more prompts. But not in this PR, just noting it down

src/session_server_ssh.c
```diff
@@ -2007,16 +1634,62 @@ nc_ssh_bind_add_hostkeys(ssh_bind sbind, struct nc_server_ssh_opts *opts)
...
         nc_timeouttime_get(&ts_timeout, opts->auth_timeout * 1000);
     }
+#if LIBSSH_0_12
+    /* Create an event loop */
+    event = ssh_event_new();
+    if (!event) {
+        ERR(session, "Failed to create SSH event.");
+        return -1;
+    }
+    ssh_event_add_session(event, session->ti.libssh.session);
```
Roytak: Check return value, i.e. `if (ssh_event_add_session(...) == SSH_ERROR) ...`

src/session_server_ssh.c
```diff
@@ -40,1900 +33,1534 @@
...
     }
+}
 
-    /* write data */
-    written += fwrite(in, 1, strlen(in), file);
+struct nc_auth_client *
+nc_ssh_find_auth_client(struct nc_server_ssh_opts *opts, const char *user, struct nc_session *session)
+{
+    struct nc_endpt *referenced_endpt;
+    uint32_t u;
```
Roytak: Should be `LY_ARRAY_COUNT_TYPE` instead of `uint32_t`

Roytak (changes requested):

Roytak (approved):

michalvasko (approved):

Issue #627, opened 2026-08-21
netopeer2-server not connecting to second client


yoousra:
Hello,

I have a test with the following scenario: 1 netconf server in front of 2 netconf clients

client 1 with IP address X
client 2 with IP address Y

netopeer2-server connects to client 1
netopeer2-server connects to client 2

callhome successfully established for both

kill the 2 sessions sucessfully

delete callhome data from sysrepo and unallocate IP addresses for both clients:
```
18:43:28.370 INFO/UE_APP-OAM/netopeer2-server(PID:2175): [INF]: LN: Session 4: Call Home client "client 1" removed, but an established session will not be terminated.
18:43:28.371 netopeer2-server(PID:2175): [INF]: LN: Session 4: Call Home client "client 1" thread exit.
```

client 2 with address Y starts listening
reconfigure callhome data in sysrepo for client 1 with remote-address X
netopeer2-server tries to connect to it:
```
18:47:44.640 netopeer2-server(PID:11205): [INF]: LN: Call Home client "client 1" endpoint "default-ssh" connecting...
18:47:44.640 netopeer2-server(PID:11205): [INF]: LN: Trying to connect via IPv4 to X:4334.
18:47:44.640 netopeer2-server(PID:11205): [INF]: LN: Successfully connected to X:4334 over IPv4.
18:47:54.641 netopeer2-server(PID:11205): [ERR]: LN: **SSH key exchange timeout.**
```
And it continues to try to connect with same error `SSH key exchange timeout`.
Maybe the SSH key exhange timeout is normal since there is no client listening with that IP address for now.

Then meanwhile, we configure client 2 with remote-address Y in sysrepo **but netopeer2-server doesn't initiate the callhome** as expected, there might be a problem here, what do you think?

Thank you for your reply.

Roytak:
Hello, what versions of libnetconf2 and netopeer2 are you using?

yoousra:
Hi,

Using the versions: netopeer2-v2.7.0-12-g2809ede and libnetconf2-v4.1.2-34-g50defb3

Thank you.

Roytak:
Should now be fixed in latest `devel` (libnetconf2 v4.5.0).

PR #630, opened 2026-08-25
Fix silently lost configuration updates and ordered-by user lists

Roytak:
Part of #627.

A configuration apply needs the config write lock, but that lock is held in read mode across entire transport handshakes, which can legitimately take longer than the 10s apply timeout (`auth-timeout` defaults to 30s and may be 0 = unbounded). The apply then timed out and failed without logging anything, so a newly configured Call Home client was never dispatched.

- log every lock timeout and wait longer than any handshake when applying
- do not redispatch Call Home clients that already have a running thread
- apply the order of the `endpoint` and `host-key` ordered-by user lists, which also fixes a crash when an entry is moved (reported as a replace)

Instead of discarding the change when a handshake is already in progress, this update forces the apply action to wait for the handshake to complete. Removing that wait entirely is the follow-up PR.

### DIFF
--- a/src/session_p.h
+++ b/src/session_p.h
@@ -147,6 +147,16 @@ extern struct nc_server_opts server_opts;
 #define NC_CONFIG_LOCK_TIMEOUT 10000
 
 /**
+ * @brief Timeout in msec for the locks acquired while applying a new configuration.
+ *
+ * A reader may hold the config_lock for the whole duration of a transport handshake (TCP connect,
+ * SSH/TLS key exchange and authentication), which is far longer than ::NC_CONFIG_LOCK_TIMEOUT.
+ * Giving up here means losing the configuration change, which the caller generally cannot recover
+ * from, so wait much longer than any handshake can take.
+ */
+#define NC_CONFIG_APPLY_LOCK_TIMEOUT 300000
+
+/**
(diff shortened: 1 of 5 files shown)


src/session_p.h
```diff
@@ -146,6 +146,16 @@ extern struct nc_server_opts server_opts;
...
 
+/**
+ * @brief Timeout in msec for the locks acquired while applying a new configuration.
+ *
+ * A reader may hold the config_lock for the whole duration of a transport handshake (TCP connect,
+ * SSH/TLS key exchange and authentication), which is far longer than ::NC_CONFIG_LOCK_TIMEOUT.
+ * Giving up here means losing the configuration change, which the caller generally cannot recover
+ * from, so wait much longer than any handshake can take.
+ */
+#define NC_CONFIG_APPLY_LOCK_TIMEOUT 300000
```
michalvasko, line 157: Always include the units of the timeout, at least in the description.

michalvasko (changes requested):

src/session_p.h
Roytak: Adjusted in the follow-up PR.

michalvasko (approved):

PR #637, opened 2026-09-04
Fix pollsession turn queue starving session addition and removal

### LINKED ISSUE #561: Failed to write reply (would block) & POLL_IO_TIMEOUT
Hello !

I'm using versions:

_netopeer2-2.2.35
sysrepo-3.3.10
libnetconf2-3.5.5
libyang-3.7.8_

It seems I sporadically encounter an error with this message:
 **[ERR]: LN: Session 1: Failed to write reply (would block)**

Note that in my case I am running netopeer2 with POLL_IO_TIMEOUT set to 300 because the host is sometimes slow.
https://github.com/search?q=repo%3ACESNET%2Fnetopeer2%20POLL_IO_TIMEOUT&type=code

I noticed, and another person as well, that it seems that increasing the POLL_IO_TIMEOUT causes all the rpcs to be delayed by the POLL_IO_TIMEOUT amount.

From my understand, this should be a timeout, meaning the maximum value before raising an error, not the time between 2 rpcs right ?

Because I tried to increase the value which shouldn't have side effect, but then I have new errors due to this "lock"
Issue seems located around this line, because removing it seems to help in our scenario, but it might no be safe:

https://github.com/CESNET/libnetconf2/blob/61fbe731908809f88a187f223d43718479a7e0da/src/session_server.c#L1410

Roytak:
The turn queue was strictly FIFO, so `nc_ps_add_session()` and the other
bounded operations queued up behind every poll thread, each holding its turn
for the whole timeout it was given so a new session went unpolled for up to
`thread_count * poll_timeout` msec.

Threads now queue up with a priority: everything except `nc_ps_poll()` only
walks the session array, so it is queued in front of the poll threads, and the
polling thread hands the turn over once it is no longer first. Queue entries
are `pthread_t`, so there are no generated IDs. The queue also grows on demand
instead of failing once `PSPOLL_THREAD_COUNT` threads share a pollsession,
which silently dropped the session being added.

Also fixes the session idle timeout being compared against the poll deadline
instead of the current time. New `test_ps_poll` covers the above.

Fixes #561


CMakeLists.txt
```diff
@@ -100,13 +100,26 @@ option(ENABLE_COMPLY_WITH_ORAN_WG11 "Comply with the O-RAN WG11 security spec (e
 option(BUILD_SHARED_LIBS "By default, shared libs are enabled. Turn off for a static build." ON)
 set(READ_INACTIVE_TIMEOUT 20 CACHE STRING "Maximum number of seconds waiting for new data once some data have arrived")
 set(READ_ACTIVE_TIMEOUT 300 CACHE STRING "Maximum number of seconds for receiving a full message")
-set(MAX_PSPOLL_THREAD_COUNT 6 CACHE STRING "Maximum number of threads that could simultaneously access a ps_poll structure")
+set(PSPOLL_THREAD_COUNT 6 CACHE STRING "Expected number of threads that simultaneously access a ps_poll structure, its queue grows past this value if needed")
```
michalvasko: Can be completely removed.

CMakeLists.txt
```diff
@@ -100,13 +100,26 @@ option(ENABLE_COMPLY_WITH_ORAN_WG11 "Comply with the O-RAN WG11 security spec (e
...
 set(TRANSPORT_HANDSHAKE_TIMEOUT 10 CACHE STRING "SSH key exchange and TLS handshake timeout in seconds")
 set(MESSAGE_MAX_SIZE 1048576 CACHE STRING "Maximum size of a message in kB")
 set(TIMEOUT_STEP 100 CACHE STRING "Number of microseconds tasks are repeated until timeout elapses")
 set(YANG_MODULE_DIR "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/yang/modules/libnetconf2" CACHE STRING "Directory where to copy the YANG modules to")
 set(CLIENT_SEARCH_DIR "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/yang/modules" CACHE STRING "Default NC client YANG module search directory")
 
+# deprecated options, kept working (with a warning) for the time being
```
michalvasko: Options can be removed, `cmake` does not fail if they are defined, they will just be ignored.

src/session_server.c
```diff
@@ -1862,154 +1862,345 @@ nc_accept_inout(int fdin, int fdout, const char *username, const struct ly_ctx *
...
     }
 
-    /* get a unique queue value (by adding 1 to the last added value, if any) */
-    if (ps->queue_len) {
-        q_last = (ps->queue_begin + ps->queue_len - 1) % NC_PS_QUEUE_SIZE;
-        *id = ps->queue[q_last] + 1;
+    if (!ps->queue_size) {
+        /* start with the configured size */
+        new_size = (NC_PS_QUEUE_SIZE > 1) ? NC_PS_QUEUE_SIZE : 2;
+    } else if (ps->queue_size > UINT8_MAX / 2) {
```
michalvasko: Why only half of `UINT8_MAX`?

src/session_server.c
```diff
@@ -1862,154 +1862,345 @@ nc_accept_inout(int fdin, int fdout, const char *username, const struct ly_ctx *
...
         ERRINT;
         return;
     }
 
-    --ps->queue_len;
-    if (found == 1) {
-        /* remove the id by moving the queue, otherwise all the values in the queue were moved */
-        ps->queue_begin = (ps->queue_begin + 1) % NC_PS_QUEUE_SIZE;
+    /* a thread is identified by its ID in the queue, so it must not be in it twice */
+    if (!nc_ps_queue_find(ps, &idx)) {
```
michalvasko: Redundant.

src/session_p.h
```diff
@@ -1227,9 +1229,14 @@ struct nc_pollsession {
...
     pthread_cond_t cond;
     pthread_mutex_t lock;
-    uint8_t queue[NC_PS_QUEUE_SIZE]; /**< round buffer, queue is empty when queue_len == 0 */
+    pthread_t *queue;                /**< round buffer with the threads waiting for their turn, the high
+                                          priority ones first, grows on demand, empty when queue_len == 0 */
+    uint8_t queue_size;              /**< allocated size of queue, 0 until the first thread queues up */
     uint8_t queue_begin;             /**< queue starts on queue[queue_begin] */
-    uint8_t queue_len;               /**< queue ends on queue[(queue_begin + queue_len - 1) % NC_PS_QUEUE_SIZE] */
+    uint8_t queue_len;               /**< queue ends on queue[(queue_begin + queue_len - 1) % queue_size] */
+    pthread_t turn_tid;              /**< thread that is currently working with the pollsession,
```
michalvasko: Seems redundant.

src/session_server.c
```diff
@@ -1862,154 +1862,345 @@ nc_accept_inout(int fdin, int fdout, const char *username, const struct ly_ctx *
...
+ * Expects @p ps->lock to be held and the queue not to be full.
+ *
+ * @param[in,out] ps Pollsession structure.
+ * @param[in] prio Whether this thread has a high priority, meaning it is added at the very
+ * beginning of the queue instead of at its end. The order in which several high priority threads
+ * that queued up at the same time get their turn is not defined, they all only walk the session
+ * array so none of them can block the others for long.
+ */
+static void
+nc_ps_queue_add(struct nc_pollsession *ps, int prio)
```
michalvasko: I like `preempt` a bit more than `prio`, more accurate.

src/session_server.c
```diff
@@ -1862,154 +1862,345 @@ nc_accept_inout(int fdin, int fdout, const char *username, const struct ly_ctx *
...
-    /* remove ourselves from the queue */
-    nc_ps_queue_remove_id(ps, id);
-    DBL(NULL, "PS 0x%p TID %lu queue: removed %u, head %u, length %u", ps, (long unsigned int)pthread_self(), id,
-            ps->queue[ps->queue_begin], ps->queue_len);
+    /* give up the turn and remove ourselves from the queue */
+    ps->turn_taken = 0;
+    nc_ps_queue_remove(ps);
 
     /* broadcast to all other threads that the queue moved */
     pthread_cond_broadcast(&ps->cond);
```
michalvasko: Knowing `pid`s now, you can instead `pthread_cond_signal()` to the next-in-queue thread. If the queue is empty, no cond signal should be needed.

src/session_server.c
```diff
@@ -1862,154 +1862,345 @@ nc_accept_inout(int fdin, int fdout, const char *username, const struct ly_ctx *
...
+        goto cleanup;
+    }
+
+    /* give up the turn and queue up again at the end, no queue slot is freed so this cannot fail */
+    ps->turn_taken = 0;
+    nc_ps_queue_remove(ps);
+    nc_ps_queue_add(ps, 0);
+
+    /* broadcast to all other threads that the queue moved */
+    pthread_cond_broadcast(&ps->cond);
```
michalvasko: Definitely more efficient to `pthread_cond_signal()` in this case.

src/session_server.c
```diff
@@ -1862,154 +1862,345 @@ nc_accept_inout(int fdin, int fdout, const char *username, const struct ly_ctx *
...
+
+    if (pthread_equal(ps->queue[ps->queue_begin], pthread_self())) {
+        /* we are still at the beginning of the queue, no one is waiting in front of us */
+        goto cleanup;
+    }
+
+    /* give up the turn and queue up again at the end, no queue slot is freed so this cannot fail */
+    ps->turn_taken = 0;
+    nc_ps_queue_remove(ps);
+    nc_ps_queue_add(ps, 0);
```
michalvasko: Why change the queue at all? Why let other threads queued after this thread preempt this thread? Just give up the turn and wait until the preempted threads finish their work, then this thread should continue.

michalvasko (changes requested):

CMakeLists.txt
Roytak: Removed from both `CMakeLists.txt`, `README.md` and `libnetconf2.pc.in`. Will need to update `netopeer2`'s cmake accordingly.

src/session_server.c
Roytak: Added an extra branch to cap at UINT8_MAX.

src/session_p.h
Roytak: Removed.

src/session_server.c
Roytak: Renamed.

src/session_server.c
Roytak: Done.

src/session_server.c
Roytak: Changed to `pthread_cond_signal` at the cost of each thread needing it's own cond.

michalvasko: Sorry, my bad, for some reason I have assumed that `pthread_cond_signal()` can wake up a specific thread based on its TID, which is not true. So please revert this change, it's not worth it with individual cond variables.

src/session_p.h
```diff
@@ -95,9 +95,15 @@ extern struct nc_server_opts server_opts;
...
 /**
- * Timeout in msec for a thread to wait for its turn to work with a pollsession structure.
+ * Initial number of threads the queue of a pollsession structure is allocated for, it grows on demand.
  */
-#define NC_PS_QUEUE_TIMEOUT 5000
+#define NC_PS_QUEUE_INIT_SIZE 6
+
+/**
+ * Timeout in msec for a thread to get the pollsession lock and its turn. The thread that has the
+ * turn may be freeing sessions in ::nc_ps_clear(), which blocks for a while on each of them.
```
michalvasko: Not a good use-case, you can always free sessions without holding the lock by first removing them, unlocking, and only then calling free. Update `nc_ps_clear()` so that it behaves this way. Then you can set this timeout to a more sensible lower value.

michalvasko (changes requested):
Changes themselves look good, just an additional change seems appropriate.

michalvasko (approved):

PR #639, opened 2026-09-07
server config FEATURE public API for hashing a password

niklas-moser:
Splits the salt generation and the crypt(3) call out of nc_server_config_add_ssh_user_password() into a public nc_server_config_hash_password(). No behaviour change.

Today the logic is only reachable through a function that builds an entire ietf-netconf-server subtree around the result. A caller that wants just the digest has to duplicate the crypt code and take on a libcrypt dependency of its own — which is what CESNET/Netopeer2#<N> would otherwise have to do.

Motivation is: O-RAN WG11 R004, and 3GPP TS 33.117 clause 4.2.3.4.3.1 behind it, don't let a deployment ship clear-text credentials in a datastore, so Netopeer2 has to hash a "$0$" password at commit time


src/server_config.h
```diff
@@ -498,6 +498,18 @@ int nc_server_config_add_ssh_user_authkey(const struct ly_ctx *ctx, const char *
...
+ * @brief Hashes a clear-text password the way the server stores it in its configuration.
+ *
+ * Produces a value usable as an 'ietf-ssh-server' hashed-password, that is a crypt(3) SHA-512
+ * digest under a freshly generated random salt ("$6$<salt>$<digest>").
+ *
+ * @param[in] password Clear-text password to hash.
+ * @param[out] hashed_password Hashed password, memory is allocated and has to be freed by the caller.
+ * @return 0 on success, non-zero otherwise.
+ */
+int nc_server_config_hash_password(const char *password, char **hashed_password);
```
Roytak, line 511: It would make more sense to me, if this API accepted the whole `crypt-hash` value (with the `$0$`) and do something only if the `$0$` is present. This would need to be documented.

niklas-moser: Makes sense! Changed it

Roytak (approved):

michalvasko (approved):

PR #640, opened 2026-09-07
session server ssh FEATURE lock an account out after repeated failed password authentication

niklas-moser:
Nothing reads ssh_auth_attempts, so password guessing is unlimited both within a connection and across them. auth-timeout bounds how long one authentication may take, not how many may be tried, and pam_faillock only ever sees the
keyboard-interactive method — accounts using hashed-password are verified by libnetconf2 itself with crypt(3).

This counts consecutive password failures per account across connections and refuses the account for NC_AUTHLOCK_TIME after NC_AUTHLOCK_MAX_FAILS. The hooks sit on the three credential checks in session_server_ssh.c that both the message-and callback-based backends funnel through (auth_password_check, kbdint_verify_passwd, pam_authenticate), so all three methods share one tally and neither dispatch file is touched. The tally is mirrored to a state file and re-read when it changes, so a lockout survives a restart and can be cleared on a running server. A session hitting NC_AUTHLOCK_SESSION_MAX_FAILS is disconnected — ssh_auth_attempts finally gets a reader.

Public key auth is deliberately not counted, which keeps a locked out deployment recoverable. TLS is unaffected.

Open question: the policy is compiled in (5 failures, 300 s, 900 s window, 6 per session, 64 accounts). ietf-netconf-server has no leaves for it and your augment carries only auth-timeout, so making it configurable means extending
libnetconf2-netconf-server.yang. If you want to make it configurable, we can change the approach there

Motivated by O-RAN WG11 R004 / 3GPP TS 33.117 4.2.3.4.3.1.


src/session_server_ssh.c
```diff
@@ -48,6 +50,403 @@
...
+
+/**
+ * @brief Record the outcome of a password authentication against an account.
+ *
+ * @param[in] session NETCONF session, for the log message.
+ * @param[in] username Account that authenticated.
+ * @param[in] success Whether it succeeded, which clears the tally.
+ */
+static void
+nc_authlock_record(struct nc_session *session, const char *username, int success)
```
Roytak, line 475: The problem I have with this is that the tally is keyed on the username alone, and the username is entirely attacker-controlled. Anyone who can reach the NETCONF port can keep `root`/`admin` permanently locked out of password auth by sending 5 bad passwords every 300 s. At minimum the lockout should be keyed by (`username`, `peer address`), and the deployment must be able to turn it off. For the peer address I believe you can use `session->host`, but it may be NULL...

src/session_server_ssh.c
```diff
@@ -48,6 +50,403 @@
...
+ * connections, and the account is refused password authentication for
+ * NC_AUTHLOCK_TIME seconds once NC_AUTHLOCK_MAX_FAILS is reached. The tally is
+ * mirrored to a state file so that restarting the server does not clear a
+ * lockout. Public key and certificate authentication are deliberately left
+ * alone: the lockout is on guessing a password, and leaving key auth open keeps
+ * a locked-out deployment recoverable.
+ */
+#define NC_AUTHLOCK_MAX_FAILS 5             /**< consecutive failures that lock an account out */
+#define NC_AUTHLOCK_TIME 300                /**< how long an account stays locked out, seconds */
+#define NC_AUTHLOCK_FAIL_INTERVAL 900       /**< failures further apart than this start a new tally, seconds */
```
Roytak: Every existing libnetconf2 user gets a behaviour change they cannot configure, disable, or even observe except through a WRN. Add an augment to the `libnetconf2-netconf-server` YANG with the leaves (e.g. `max_fails`, `lock_time`, `window` probably somewhere under SSH's `client-authentication`) and default to disabled, so this is opt-in.

src/session_server_ssh.c
```diff
@@ -48,6 +50,403 @@
...
+         * known to the endpoint reach the authentication dispatch, so this bound is about memory
+         * rather than about an attacker flooding the tally with invented names. */
+        slot = 0;
+        for (i = 1; i < authlock.entry_count; ++i) {
+            if (authlock.entries[i].last_fail < authlock.entries[slot].last_fail) {
+                slot = i;
+            }
+        }
+        free(authlock.entries[slot].username);
+    }
```
Roytak, line 429: You evict the entry with the oldest `last_fail` when the table is full here, without regard for whether that entry is currently locked. I think you should never evict an entry with `locked_until` > `now`, prefer evicting expired/unlocked entries and if none are available, refuse to add rather than dropping a lockout.

src/session_server_ssh.c
```diff
@@ -48,6 +50,403 @@
...
+ *
+ * @return Path from $NC_AUTHLOCK_FILE, or the compiled-in default.
+ */
+static const char *
+nc_authlock_path(void)
+{
+    const char *path = getenv("NC_AUTHLOCK_FILE");
+
+    return (path && path[0]) ? path : NC_AUTHLOCK_FILE;
+}
```
Roytak, line 147: Nothing in the build creates `/var/lib/netconf-authlock/`, so on a normal install every open() fails, a single WRN is logged, and the persistence half of the feature is silently absent.

Make the default path a CMake option (off by default), expose a setter in the public API, drop the `getenv`, and create the directory (or log loudly and disable persistence).

src/session_server.c
```diff
@@ -1614,6 +1619,9 @@ nc_server_destroy(void)
         ERR(NULL, "%s: failed to stop certificate expiration notification thread.", __func__);
         goto cleanup;
     }
+
+    /* free the password authentication lockout tally, the state file it mirrors is kept */
+    nc_server_ssh_authlock_free();
```
Roytak: It sits before `nc_server_ch_threads_destroy()` so Call Home and accept threads are still authenticating clients. Move the call to the end of the teardown, after all threads are joined.

src/session_server_ssh.c
```diff
@@ -501,6 +928,9 @@ nc_server_ssh_pam_authenticate(struct nc_session *session, const char *username,
         ERR(NULL, "PAM error occurred (%s).", pam_strerror(pam_h, ret));
     }
     free(pam_config_name);
+
+    nc_authlock_record(session, username, (ret == PAM_SUCCESS) ? 1 : 0);
+
```
Roytak: Should record only if `pam_authenticate()` fails with an auth failure retcode.

src/session_server_ssh.c
```diff
@@ -48,6 +50,403 @@
...
+ * how an operator clears a lockout early: remove the file, or edit the account out of it. Writes
+ * this code made itself are not read back.
+ */
+static void
+nc_authlock_sync(void)
+{
+    struct stat st;
+
+    if (stat(nc_authlock_path(), &st)) {
+        /* no state file, so nothing is locked out; it was either never written or removed to clear
```
Roytak: Should probably check for just `ENOENT` based on the comment?

src/session_server_ssh.c
```diff
@@ -48,6 +50,403 @@
...
+    }
+
+    fclose(f);
+}
+
+/**
+ * @brief Write the tally to the lockout state file. Expects the lock to be held.
+ */
+static void
+nc_authlock_store(void)
```
Roytak: Each individual failure re serialises the entire table and does open/write/rename, with `authlock.lock`, so all concurrent authentications across all endpoints need to wait.  I think persist only when `locked_until` actually changes (plus on clear).

src/session_server_ssh.c
```diff
@@ -160,6 +559,20 @@ nc_server_ssh_auth_attempt_failed(struct nc_session *session)
     ++session->opts.server.ssh_auth_attempts;
     VRB(session, "Failed user \"%s\" authentication attempt (#%d).",
             session->username ? session->username : "unknown", session->opts.server.ssh_auth_attempts);
+
+    if (session->opts.server.ssh_auth_attempts < NC_AUTHLOCK_SESSION_MAX_FAILS) {
+        return;
+    }
```
Roytak: `nc_server_ssh_auth_attempt_failed()` is called for every rejected pubkey too, so a client whose ssh-agent offers more than six keys fails here. I think that's okay, but the `max-fails` value needs to be configurable through YANG.

src/session_server_ssh.c
```diff
@@ -160,6 +559,20 @@ nc_server_ssh_auth_attempt_failed(struct nc_session *session)
...
+
+    /* the per-session cap, which bounds what a single connection may try; the accept loops end on a
+     * session that is no longer connected */
+    if (NC_SESSION_STATUS_GET(session) != NC_STATUS_INVALID) {
+        ERR(session, "Too many failed authentication attempts (%d) in a single session, disconnecting.",
+                session->opts.server.ssh_auth_attempts);
+        NC_SESSION_STATUS_SET(session, NC_STATUS_INVALID);
+        ATOMIC_STORE_RELAXED(session->term_reason, NC_SESSION_TERM_OTHER);
```
Roytak: Use `NC_SESSION_TERM_REASON_SET`.

src/session_server_ssh.c
```diff
@@ -48,6 +50,403 @@
...
+            entry->fails = 0;
+        }
+
+        ++entry->fails;
+        entry->last_fail = now;
+
+        if (entry->fails >= NC_AUTHLOCK_MAX_FAILS) {
+            entry->locked_until = now + NC_AUTHLOCK_TIME;
+            WRN(session, "User \"%s\" locked out of password authentication for %d s after %u consecutive "
+                    "failed attempts.", username, NC_AUTHLOCK_TIME, entry->fails);
```
Roytak: Use `PRIu32` instead of `%u` (3 occurences I believe).

src/session_server_ssh.c
```diff
@@ -48,6 +50,403 @@
...
+    /* remember which revision of the file this tally is, so that a change made to it from the
+     * outside is noticed */
+    if (!fstat(fileno(f), &st)) {
+        authlock.loaded_ino = st.st_ino;
+        authlock.loaded_mtim = st.st_mtim;
+    }
+
+    /* the username is the rest of the line so that it may contain spaces */
+    while ((authlock.entry_count < NC_AUTHLOCK_MAX_ENTRIES) && fgets(line, sizeof line, f)) {
+        if (sscanf(line, "%u %lld %lld %n", &fails, &locked_until, &last_fail, &offset) != 3) {
```
Roytak: Should use `SCNu32` and `uint32_t fails`.

src/session_server_ssh.c
```diff
@@ -48,6 +50,403 @@
...
+
+static struct {
+    pthread_mutex_t lock;
+    struct nc_authlock_entry entries[NC_AUTHLOCK_MAX_ENTRIES];
+    uint32_t entry_count;
+    int loaded;                 /**< whether the state file was read already */
+    ino_t loaded_ino;           /**< inode of the state file revision the tally came from */
+    struct timespec loaded_mtim; /**< mtime of the state file revision the tally came from */
+    int store_failed;           /**< whether a failure to write the state file was logged already */
+} authlock = {.lock = PTHREAD_MUTEX_INITIALIZER};
```
Roytak: Should be moved to `session_p.h` along with any needed macros.

src/session_server_ssh_wrapper.h
```diff
@@ -339,6 +339,13 @@ int nc_server_ssh_compare_password(const char *stored_pw, const char *received_p
  */
 void nc_server_ssh_auth_attempt_failed(struct nc_session *session);
 
+/**
+ * @brief Free the password authentication lockout tally.
+ *
+ * The state file it mirrors is kept, so a lockout survives the server being restarted.
+ */
+void nc_server_ssh_authlock_free(void);
```
Roytak: Belongs to `session_p.h` imo.

Roytak (changes requested):
Thanks for the contribution and sorry for the delay! I think this feature is very useful, but will need a bit more work - mainly making this optional (off by default) and configurable. Also could add at least one test e.g. for locking an account out after N attempts.

src/session_server_ssh.c
niklas-moser: Agreed, this is a real denial of service and I will key the tally on (username, peer address), but I would keep the per-account component rather than replace it, since peer-only keying leaves a distributed guesser unbounded and `3GPP TS 33.117 4.2.3.4.3.1` asks for the account-level lockout specifically.

src/session_server_ssh.c
niklas-moser: I will add the augment to `libnetconf2-netconf-server` under SSH client-authentication with the policy leaves and the feature disabled by default.

src/session_server_ssh.c
niklas-moser: Makes sense, thanks

niklas-moser:
Thank you @Roytak . I added two commits for you to review regarding the requested changes. If you want I can squash them after re-review, or you can do it.

src/session_server_ssh.c
```diff
@@ -48,6 +51,541 @@
...
+        if (strcmp(host, "-")) {
+            entry->host = strdup(host);
+            if (!entry->host) {
+                free(entry->username);
+                break;
+            }
+        }
+        entry->fails = fails;
+        entry->locked_until = locked_until;
+        entry->last_fail = last_fail;
```
Roytak, line 208: Could probably add a sanity check just in case, something like:
`locked_until > now + NC_AUTHLOCK_TIME` -> `now + NC_AUTHLOCK_TIME`
`last_fail > now` -> `now`

modules/libnetconf2-netconf-server@2026-09-18.yang
```diff
@@ -159,6 +163,67 @@ module libnetconf2-netconf-server {
...
+        type uint16 {
+          range "1..max";
+        }
+        default 5;
+        description
+          "Number of consecutive failed password authentications that lock the account out.";
+      }
+
+      leaf lock-time {
+        type uint16;
```
Roytak, line 211: Should also add `range "1..max"`.

modules/libnetconf2-netconf-server@2026-09-18.yang
```diff
@@ -159,6 +163,67 @@ module libnetconf2-netconf-server {
...
+        default 300;
+        units "seconds";
+        description
+          "How long the account stays locked out of password authentication.";
+      }
+
+      leaf fail-window {
+        type uint16;
```
Roytak, line 219: Should also add `range "1..max"`.

src/session_server_ssh.c
```diff
@@ -48,6 +51,541 @@
...
+    pthread_mutex_t lock;
+    struct nc_authlock_entry entries[NC_AUTHLOCK_MAX_ENTRIES];
+    uint32_t entry_count;
+    char *path;                 /**< state file the tally is mirrored to, NULL to keep it in memory only */
+    int path_set;               /**< whether @p path was resolved already */
+    int loaded;                 /**< whether the state file was read already */
+    ino_t loaded_ino;           /**< inode of the state file revision the tally came from */
+    struct timespec loaded_mtim; /**< mtime of the state file revision the tally came from */
+    int store_failed;           /**< whether the state file turned out to be unusable, disables persistence */
+} authlock = {.lock = PTHREAD_MUTEX_INITIALIZER};
```
Roytak, line 79: For consistency, give this struct a name, doxygen docs,  and move its definition to `session_p.h`. Keep the global variable definition here afterwards.

src/session_server_ssh.c
```diff
@@ -48,6 +51,541 @@
...
+                    (authlock.entries[i].last_fail < authlock.entries[slot].last_fail)) {
+                slot = i;
+            }
+        }
+        if (slot == NC_AUTHLOCK_MAX_ENTRIES) {
+            /* every tracked pair is locked out, so refuse to track another one rather than drop a
+             * lockout; the pairs that are locked out keep being refused either way */
+            WRN(NULL, "Authentication failure tally full of locked out users, not counting the "
+                    "failures of user \"%s\" for now.", username);
+            return NULL;
```
Roytak, line 409: This still looks easily abusable. I would recommend something along:
- add `NC_AUTHLOCK_MAX_ENTRIES_PER_HOST` and set to e.g. 8
- add `host_count = ` after `nc_authlock_find` misses in this function
- change `authlock.entry_count < NC_AUTHLOCK_MAX_ENTRIES` to `(authlock.entry_count < NC_AUTHLOCK_MAX_ENTRIES) && (host_count < NC_AUTHLOCK_MAX_ENTRIES_PER_HOST)`
- in the for loop above add another if case, e.g. `if (!nc_authlock_same_host(authlock.entries[i].host, host))` and `continue` in that case.
- instead of WRN and fail, evict the last/min last fail

Also in this file on line :1050, drop the `PAM_USER_UNKNOWN` - mirrors the shadow/local-users path. Also you can possibly just prune the table while walking it in the for above, i.g. entries with `locked_until <= now` and `last_fail` older than `fail_window` are dead weight.

modules/libnetconf2-netconf-server@2026-09-18.yang
```diff
@@ -159,6 +163,67 @@ module libnetconf2-netconf-server {
...
+         anyone else. The tally is shared by every password-based method the server offers, which
+         are the configured password, keyboard-interactive and PAM. Public key authentication is
+         not counted and never refused, so a locked out deployment stays reachable.
+
+         The policy configured here is the one of the endpoint the connection arrived on; it is not
+         re-evaluated when authentication falls through to a referenced endpoint.
+
+         The tally is kept in memory. It additionally survives a server restart if the server was
+         built with the AUTHLOCK_FILE option or the application set a path with the
+         nc_server_ssh_set_authlock_path() API call.";
```
Roytak, line 196: Overall pretty long. No need to mention implementation details in the YANG description, so the last paragraph + (username, client address) part can be dropped. Other parts possibly too.
It should plainly metion something along: "it's an account-lockout control for 3GPP 33.117 compliance, that the brute-force rate limit is `max-auth-attempts` + `auth-timeout`, and that enabling `lockout` without `max-auth-attempts` still enables one connection to make an unlimited amount of attempts".

Roytak (changes requested):
Needs some minor changes + 1 bigger one to `nc_authlock_get`. Also needs a rebase on top of `devel` due to a conflict. Otherwise okay.

PR #641, opened 2026-09-10
Fix concurrent nc_client_init() and nc_server_init()

Roytak:
`nc_client_init()` and `nc_server_init()` initialized the same process-global libraries without any lock, so doing both at once from two threads raced inside them.

Both sides now go through `nc_global_init()`/`nc_global_destroy()`, which take a lock and track which side initialized the libraries. They are therefore also initialized only once and destroyed only when neither side uses them any longer, so `nc_client_destroy()` no longer tears down the crypto state of a running server and vice versa.

`test_thread_messages` also never destroyed either side.

Seen in: https://github.com/CESNET/libnetconf2/actions/runs/34359810512


src/session.c
```diff
@@ -53,6 +53,16 @@
 /* in milliseconds */
 #define NC_CLOSE_REPLY_TIMEOUT 200
 
+#ifdef NC_ENABLED_SSH_TLS
+
+/* lock protecting the global initialization state of the libraries used by libnetconf2 */
+static pthread_mutex_t glob_init_lock = PTHREAD_MUTEX_INITIALIZER;
+
+/* whether the client/server side has the libraries initialized, protected by glob_init_lock */
+static int glob_init_client, glob_init_server;
```
michalvasko: Seems cleaner to not add unnecessary global variables and rather use `static` local ones. That will lead to a some functionality together in a single function but that should be fine.

michalvasko (changes requested):

src/session.c
Roytak: Changed.

michalvasko (approved):

